Issue #25, opened 2026-01-03
Account Creation: UI doesn't match code expectations


stephondoestech:
On the initial account creation screen the UI says a password needs to be at least 8 characters long, but if you try one that length you get a UI error back that says it needs to be 12 characters long. Not a huge deal, but wanted to mention it. 

<img width="1101" height="796" alt="Image" src="https://github.com/user-attachments/assets/b1a66f29-d270-4fe2-805a-2149a0344588" />

stephondoestech:
@Pedro-Revez-Silva small fix here. Just text changes

PR #26, opened 2026-01-03
Account Creation: Password Text Matching 12 characters

stephondoestech:
On the initial account creation screen I updated the text so it says 12 characters to match the backend expectations so users won't get an error for setting a password that is too short. 

#25


Pedro-Revez-Silva:
The minlength attribute that also needed fixing in a couple fields, pushed all the changes to main.
Thank you for the contribution.

PR #129, opened 2026-02-03
Add basic API

xarnuul:
- Adds a basic API with a single endpoint for creating new users
- Adds an API token to settings


CodingByHarry:
@Pedro-Revez-Silva This would be helpful and was actually something I was looking to add. It looks good to me :) Anything blocking this from being merged that I can work on?

PR #219, opened 2026-04-12
feat: add Z-Library as ebook search and download source

Thoroslives:
## Why

Anna's Archive is a great ebook source but the download path has a high barrier to entry - it requires a paid API key ($2+/mo) and a FlareSolverr/Byparr instance to bypass DDoS-Guard on the search pages. For users who just want automated ebook downloads without that infrastructure, Z-Library's eAPI provides a free alternative that only needs account credentials. Pure JSON over HTTP, no browser automation.

This doesn't replace the Anna's Archive integration. Ebook source priority is:

1. **Anna's Archive** - used when AA API key is configured (paid, fastest, broadest coverage)
2. **Z-Library** - used when Z-Library credentials are configured and no AA API key is present (free, only needs email/password)
3. **Neither** - skipped if no credentials for either are set

Prowlarr/Jackett indexers always run alongside whichever ebook source is active.

## Summary
- New `ZLibraryClient` service: auth, search, and download via Z-Library eAPI
- Search via `POST /eapi/book/search` with file type and language filters
- Download via `GET /eapi/book/{id}/{hash}/file` with direct CDN URLs
- Multi-domain login fallback with 30-minute token caching
- Z-Library results bypass seeder threshold in auto-select (direct downloads, same as AA)
- Post-download file verification rejects HTML error pages and corrupt files
- New admin settings: `zlibrary_email`, `zlibrary_password`
- Purely additive - no changes to existing AnnaArchiveClient or other services

## Testing
- `bundle exec rails test`
- 14 new tests across ZLibraryClient, SearchJob, DownloadJob, AutoSelectService

### DIFF
--- a/app/jobs/download_job.rb
+++ b/app/jobs/download_job.rb
@@ -35,6 +35,8 @@ class DownloadJob < ApplicationJob
       # Handle Anna's Archive downloads differently
       if search_result.from_anna_archive?
         handle_anna_archive_download(download, search_result)
+      elsif search_result.from_zlibrary?
+        handle_zlibrary_download(download, search_result)
       else
         handle_standard_download(download, search_result)
       end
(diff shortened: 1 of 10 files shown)


Pedro-Revez-Silva:
Good feature, but I didn't like the implementation. 
Done in PR #222

Thoroslives:
What was it about the implementation you didn't like? 
Theres a couple more PR's in the works and id like to work to mirror your preferred convention to reduce workload on yourself.

Thoroslives:
> What was it about the implementation you didn't like? Theres a couple more PR's in the works and id like to work to mirror your preferred convention to reduce workload on yourself.

I think i see looking at 222

PR #224, opened 2026-04-12
feat: send attention webhook for all mark_for_attention events

Thoroslives:
## Summary
- dispatch `request_attention` webhook after every `mark_for_attention!` call across SearchJob, DownloadJob, and DownloadMonitorJob
- only PostProcessingJob previously fired this notification, so 19 other call sites silently flagged requests without sending a webhook
- users with webhook/ntfy integrations now get notified immediately instead of polling the UI

## Motivation

The `request_attention` event type, the webhook delivery pipeline, and the `webhook_events` default all already exist. The gap was that most jobs called `mark_for_attention!` without also dispatching `NotificationService.request_attention`. This matches the pattern PostProcessingJob already uses (lines 40-41).

I considered moving the dispatch into `mark_for_attention!` itself to avoid this gap in the future, but kept it as explicit calls to match the existing PostProcessingJob pattern. Happy to refactor if you prefer it centralized.

## Testing
- `bundle exec rails test`
- 13 new tests: notification dispatch assertions + defensive tests confirming jobs still work when outbound delivery fails

### DIFF
--- a/app/jobs/download_job.rb
+++ b/app/jobs/download_job.rb
@@ -33,6 +33,7 @@ class DownloadJob < ApplicationJob
       track_request_event(download.request, "dispatch_failed", download: download, message: "No search result selected for download", level: :error)
       download.update!(status: :failed)
       download.request.mark_for_attention!("No search result selected for download")
+      NotificationService.request_attention(download.request)
       return
     end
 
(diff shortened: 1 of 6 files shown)


Pedro-Revez-Silva:
I had to refactor the attention behaviour and added this.

PR #225, opened 2026-04-13
feat: add indexer preference for download client routing

Thoroslives:
## Summary
- add optional indexer preference to download clients so users can route downloads from specific indexers to specific clients
- when a search result comes from a preferred indexer, that client is selected regardless of priority order, with automatic fallback to priority-based selection if the preferred client is unavailable
- add a dynamic checkbox picker on the download client form that fetches available indexers from the active indexer provider, with text input fallback when no provider is configured

## Motivation
Users running both a debrid client (Decypharr) and a real torrent client (qBittorrent) currently cannot route private tracker results to qBittorrent while keeping public tracker results on Decypharr. The only routing available is priority-based, which applies the same client to all downloads regardless of source.

## Scope
This is a single feature (indexer-based client routing) that touches backend, UI, and tests together. The 11 files break down as: 1 migration, 3 app changes (model, service, controller), 2 UI (form partial, Stimulus controller), 1 route, 1 schema, 3 test files. Happy to split backend and UI into separate PRs if preferred.

## Changes
- add `preferred_indexers` string column to `download_clients` (comma-separated indexer names)
- add `preferred_for_indexer?` and `preferred_indexer_list` methods to `DownloadClient`
- add cross-client validation preventing the same indexer from being assigned to multiple clients
- update `DownloadClientSelector#select` to check indexer preference before priority fallback
- add `available_indexers` collection endpoint returning indexer names with assignment info
- add Stimulus controller with checkbox UI (DOM construction, no innerHTML) and graceful degradation
- hide indexer preference section for usenet client types
- show which indexers are assigned to other clients (disabled, grayed out with client name)

## Testing
- 5 new selector tests: preference match, fallback on no match, fallback on connection failure, case-insensitive, nil indexer
- 7 new model tests: matching, blank handling, whitespace, validation, self-update, assignments map
- 3 new controller tests: available_indexers endpoint, create/update persistence
- full suite: 926 runs, 0 errors, 2 pre-existing failures (UpdateCheckerServiceTest)

### DIFF
--- a/app/models/download_client.rb
+++ b/app/models/download_client.rb
@@ -22,6 +22,7 @@ class DownloadClient < ApplicationRecord
     numericality: { only_integer: true, greater_than: 0 }
   validates :torrent_verification_wait_time,
     numericality: { only_integer: true, greater_than_or_equal_to: 0 }
+  validate :preferred_indexers_not_claimed_by_other_clients
 
   scope :enabled, -> { where(enabled: true) }
   scope :by_priority, -> { order(priority: :asc) }
@@ -67,11 +68,46 @@ class DownloadClient < ApplicationRecord
     names.any? { |name| name.casecmp(indexer_name.strip) == 0 }
   end
 
+  def preferred_indexer_list
+    return [] if preferred_indexers.blank?
+
+    preferred_indexers.split(",").map(&:strip).reject(&:blank?)
+  end
+
+  class << self
+    def indexer_assignments(exclude_client_id: nil)
+      scope = preferred_indexers_present
+      scope = scope.where.not(id: exclude_client_id) if exclude_client_id
+      scope.each_with_object({}) do |client, map|
+        client.preferred_indexer_list.each { |name| map[name.downcase] = client.name }
+      end
+    end
+
+    private
+
+    def preferred_indexers_present
+      where.not(preferred_indexers: [nil, ""])
+    end
+  end
+
   def requires_authentication?
     qbittorrent? || decypharr? || nzbget? || deluge? || transmission?
   end
 
   def qbittorrent_compatible?
     qbittorrent? || decypharr?
+  end
+
+  private
+
+  def preferred_indexers_not_claimed_by_other_clients
+    return if preferred_indexers.blank?
+
+    assignments = self.class.indexer_assignments(exclude_client_id: id)
+    conflicts = preferred_indexer_list.select { |name| assignments.key?(name.downcase) }
+    return if conflicts.empty?
+
+    conflict_details = conflicts.map { |name| "#{name} (#{assignments[name.downcase]})" }
+    errors.add(:preferred_indexers, "contains indexers already assigned to other clients: #{conflict_details.join(', ')}")
   end
 end
--- a/app/models/download_client.rb
+++ b/app/models/download_client.rb
@@ -83,7 +83,7 @@ class DownloadClient < ApplicationRecord
 
   class << self
     def indexer_assignments(exclude_client_id: nil)
-      scope = where.not(preferred_indexers: [nil, ""])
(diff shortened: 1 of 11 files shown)


app/models/download_client.rb
```diff
@@ -60,11 +61,47 @@ def usenet_client?
...
   def requires_authentication?
     qbittorrent? || decypharr? || nzbget? || deluge? || transmission?
   end
 
   def qbittorrent_compatible?
     qbittorrent? || decypharr?
   end
+
+  class << self
+    def indexer_assignments(exclude_client_id: nil)
```
Pedro-Revez-Silva, line 85: `indexer_assignments` treats every client with a non-blank `preferred_indexers` as an active claimant. That means a `sabnzbd`/`nzbget` record can reserve torrent indexers even though the UI hides this field for usenet clients and the selector only uses torrent clients. I reproduced this on the PR branch in `RAILS_ENV=test`: saving `preferred_indexers: "MyAnonaMouse"` on a `sabnzbd` client makes a qBittorrent client with the same preference invalid. This should either ignore non-torrent clients here or clear `preferred_indexers` whenever a client is switched to a usenet type.

app/models/download_client.rb
```diff
@@ -60,11 +61,47 @@ def usenet_client?
...
+      scope = scope.where.not(id: exclude_client_id) if exclude_client_id
+      scope.each_with_object({}) do |client, map|
+        client.preferred_indexer_list.each { |name| map[name.downcase] = client.name }
+      end
+    end
+  end
+
+  private
+
+  def preferred_indexers_not_claimed_by_other_clients
```
Pedro-Revez-Silva, line 96: This validation relies on `indexer_assignments`, which currently includes disabled clients as well. In practice, disabling a client does not release its preferred indexers even though `DownloadClientSelector` only considers `enabled.by_priority` clients. I reproduced this on the PR branch in `RAILS_ENV=test`: a disabled qBittorrent client with `preferred_indexers: "MyAnonaMouse"` prevents an enabled client from saving the same assignment. The validation should either ignore disabled clients or disabling a client should clear these preferences.

Pedro-Revez-Silva (changes requested):
Two behavioral issues need follow-up before this is safe to merge: non-torrent/hidden-state clients can still reserve preferred indexers, and disabled clients still block those assignments even though selector logic ignores them.

app/models/download_client.rb
Thoroslives: Good catch. Scoped `indexer_assignments` to `enabled.torrent_clients` so usenet clients are excluded from the query entirely. This matches what the selector does - it only considers `torrent_clients.enabled.by_priority`. Added a test that confirms a sabnzbd client with preferred_indexers set doesn't appear in assignments and doesn't block torrent clients from claiming the same indexer.

app/models/download_client.rb
Thoroslives: Fixed in the same change - `indexer_assignments` now scopes to `enabled.torrent_clients`, so disabled clients are excluded from both the assignment map and the validation. Added tests for both: disabled clients don't appear in assignments, and an enabled client can claim an indexer that a disabled client already has set.

Thoroslives:
Apologies for not pinging when I pushed the fix. `fix: scope indexer_assignments to enabled torrent clients` (Apr 13) addresses both points: non-torrent/hidden-state clients no longer reserve indexers, and disabled clients no longer block assignments. Should have dropped a note at the time to say so.

Rebased onto main just now to clear the `db/schema.rb` conflict (mechanical, just the version bump plus the new `series_position` and `preferred_indexers` columns on different tables). Ready for re-review whenever you have time.

Pedro-Revez-Silva:
Implemented in a different way. Should be working now if routing rules are configured.

PR #230, opened 2026-04-15
Add setting to disable Z-Library download host validation

joopdo:
Z-Library frequently serves file downloads from CDN domains that don't match the configured API domain. This causes downloads to fail with 'download URL outside the configured host family' errors.

Add a zlibrary_strict_host_check boolean setting (Admin > Settings > Z-Library) that defaults to true (preserving current behavior). When disabled, the download URL is still validated for scheme and presence of a host, but the domain family check is skipped.

Fixes both validation points: ZLibraryClient#download_host_allowed? and DownloadJob#zlibrary_download_host_allowed?.

### DIFF
--- a/app/jobs/download_job.rb
+++ b/app/jobs/download_job.rb
@@ -268,6 +268,8 @@ class DownloadJob < ApplicationJob
   end
 
   def zlibrary_download_host_allowed?(host)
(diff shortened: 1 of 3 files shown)


Pedro-Revez-Silva:
Solved in release 0.16.0

Issue #266, opened 2026-05-10
Can't connect to Qbittorrent


VespeiProjects:
When I execute inside the docker container
`curl -i -X POST -d "username=admin&password=(PASSWORD)" http://192.168.1.34:49893/api/v2/auth/login`

I get response
```
HTTP/1.1 204 OK
connection: keep-alive
content-length: 0
content-security-policy: default-src 'self'; style-src 'self' 'unsafe-inline'; img-src 'self' data:; script-src 'self' 'unsafe-inline'; object-src 'none'; form-action 'self'; frame-src 'self' blob:;
cross-origin-opener-policy: same-origin
date: Sun, 10 May 2026 04:12:38 GMT
referrer-policy: same-origin
set-cookie: QBT_SID_49893=7zzuIRM8LuHfT2MsSu/rNKql+FGY9lGh; HttpOnly; expires=Sun, 10-May-2026 05:12:38 GMT; path=/
x-content-type-options: nosniff
x-xss-protection: 1; mode=block
```
Inside Shelfarr when I add http://192.168.1.34:49893 with the same credentials I get a failed to connect.

ganievs:
Hi there!
Seems modern qBittorrent versions return **HTTP 204** with an empty body on successful login and session cookie names is port-suffixed or prefixed like `QBT_SID_49893`, `QBT_SID`

ganievs:
Both issues is related to qBittorrent v5.2.0 API changes:

HTTP 204 on login - [HA #169721](https://github.com/home-assistant/core/issues/169721)
suffixed cookie -[qBittorrent #18329](https://github.com/qbittorrent/qBittorrent/issues/18329)

Pedro-Revez-Silva:
Should be fixed in v0.19.1

PR #269, opened 2026-05-12
fix: support newer auth response and SID cookie variants

ganievs:
solves #266


ganievs:
Cheсked on qbittorent-nox v5.2.0.

PR #290, opened 2026-05-28
Added bulk upload funcationality and directory monitoring Docker mount.

### LINKED ISSUE #286: [Feature Request] Monitored folder / Bulk Upload
This app is perfect for my needs, the only thing I wish was easier is importing a whole library.

Adding a folder that could be monitored (drop box for new downloads, temporary share into friends' servers) for new works and importing them one by one according to your rules.

The upload feature is a great implementation of this, but one file at a time is tedious. Multiselect in the file dialog or even "add another file" button to queue another afterwards would be lovely.

jreidray:
Resolves #286 

Enables multi-file uploads
Extends Uploads database to mark file path
Adds /watch volume and timer that scans directory for new files
New files added to Uploads queue for processing
Files in Uploads can now be Pending, Processing, Success, or Failed

These changes met my needs and tests ran successfully. Code generated using Gemini CLI

### DIFF
--- a/app/controllers/uploads_controller.rb
+++ b/app/controllers/uploads_controller.rb
@@ -18,12 +18,24 @@ class UploadsController < ApplicationController
   end
 
   def create
-    result = UploadCreator.call(user: Current.user, uploaded_file: params[:file], request: @request)
+    uploaded_files = Array(params[:file])
+    results = uploaded_files.map do |file|
+      UploadCreator.call(user: Current.user, uploaded_file: file, request: @request)
+    end
 
-    if result.success?
-      redirect_to upload_success_location, notice: result.notice
+    success_count = results.count(&:success?)
+    failure_count = results.count { |r| !r.success? }
+
+    if success_count > 0 && failure_count == 0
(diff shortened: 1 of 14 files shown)


Pedro-Revez-Silva:
What is the use case of putting files in a /watch folder instead of uploading directly?

jreidray:
Large existing collections (gigs, terabytes) can be churned through in the background instead of having to push that many files over a network connection. Would allow deep folder structures with inconsistent organization to be handled rather than hunting down individual files in a browser.

Pedro-Revez-Silva:
> Large existing collections (gigs, terabytes) can be churned through in the background instead of having to push that many files over a network connection. Would allow deep folder structures with inconsistent organization to be handled rather than hunting down individual files in a browser.

Thanks for explaining the use case. This is not something I intend to support in Shelfarr.

Shelfarr’s scope is requests and acquisition, not bulk library migration or library management. It is already aware of the library through integrations with library managers and that is where knowledge of existing library contents should come from. I use Audiobookshelf, but I’m open to supporting other library managers as integrations.

For the case you described, where the files already exist on the same server or network, the better approach is to copy or move them into the library folder and let the library manager handle scanning/importing them.

PR #323, opened 2026-06-23
feat(settings): allow any setting to be overridden from the environment

coffbr01:
## What

Adds a config-as-code surface so any setting can be supplied from the environment instead of (only) the admin UI. Setting `SHELFARR_SETTING_<KEY>` (uppercased setting key) overrides the stored value of `<key>`, cast to its declared type.

```yaml
environment:
  - SHELFARR_SETTING_OIDC_ENABLED=true
  - SHELFARR_SETTING_OIDC_ISSUER=https://auth.example.com/application/o/shelfarr/
  - SHELFARR_SETTING_OIDC_CLIENT_ID=shelfarr
  - SHELFARR_SETTING_OIDC_CLIENT_SECRET=...   # from a secret store
  - SHELFARR_SETTING_WEBHOOK_ENABLED=true
  - SHELFARR_SETTING_WEBHOOK_URL=http://my-listener:9000/
```

## Why

Today every setting (including OIDC issuer/client-id/**client-secret** and the outbound webhook config) lives only in the app's SQLite `settings` table, written through the admin UI. For declarative / GitOps deployments (Kubernetes, Compose-with-secrets, Kamal) that has two pain points:

- **Secrets in the DB.** The OIDC client secret can't be injected from a secret manager; it has to be typed into the UI and then lives in the database.
- **Lost on a wipe/restore.** A fresh or restored DB silently drops the configuration, so auth + integrations have to be re-entered by hand.

This makes the environment the source of truth and the database a cache: an env-managed key is reconciled on every boot, so it survives a DB wipe/restore with no manual re-entry.

## How

- `SettingsService.get` consults `SHELFARR_SETTING_<KEY>` first, reusing the `Setting` model's type casting so booleans (`true`/`false`), integers, and JSON-array values parse exactly like a stored value. Because OIDC already reads live via `SettingsService.get`, this needs no boot ordering or restart.
- New helpers: `env_managed?`, `env_managed_keys`, `env_override_name`.
- **Admin → Settings** renders an env-managed key as a single read-only row, with the controlling variable behind a hover tooltip on a lock icon. The settings controller refuses to persist env-managed keys (so a stray POST can't write a shadowed value).
- Keys you don't set in the environment stay fully UI-managed, exactly as before — this is purely additive and opt-in.

## Tests

- Added `SettingsServiceTest` coverage: precedence over a stored value, supplying a value with no DB row (the wipe/restore case), boolean/integer/json casting, disabling a DB-enabled setting, `oidc_configured?` / `configured?` honoring env with no rows, and `env_managed_keys` / `env_override_name`.
- `bin/rails test test/services/settings_service_test.rb test/controllers/admin/settings_controller_test.rb` → 123 runs, 0 failures.

## Notes

- `DISABLE_AUTH` precedent already existed for one setting; this generalizes the idea to all of them under a consistent `SHELFARR_SETTING_*` prefix.
- Documented under "Settings as environment variables (config-as-code)" in the README.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,33 @@ environment:
   - RAILS_RELATIVE_URL_ROOT=/shelfarr
 ```
 
+#### Settings as environment variables (config-as-code)
+
+Any setting from **Admin → Settings** can also be supplied from the environment,
+which is useful for declarative/GitOps deployments and for keeping secrets out of
+the app database. Set `SHELFARR_SETTING_<KEY>` (the uppercased setting key) and it
+overrides the stored value:
+
+```yaml
+environment:
+  # OIDC, sourced from the environment instead of the admin UI
+  - SHELFARR_SETTING_OIDC_ENABLED=true
+  - SHELFARR_SETTING_OIDC_ISSUER=https://auth.example.com/application/o/shelfarr/
+  - SHELFARR_SETTING_OIDC_CLIENT_ID=shelfarr
+  - SHELFARR_SETTING_OIDC_CLIENT_SECRET=...   # inject from a secret store
+  # Outbound webhook
+  - SHELFARR_SETTING_WEBHOOK_ENABLED=true
+  - SHELFARR_SETTING_WEBHOOK_URL=http://my-listener:9000/
+```
+
+Semantics: **the environment is the source of truth and the database is a cache.**
+A managed key shows up read-only in the settings page (with the controlling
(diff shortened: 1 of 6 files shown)


Pedro-Revez-Silva:
Implemented in a different PR, same feature, slightly different approach.

PR #340, opened 2026-07-06
[codex] Improve indexer search broadening

### LINKED ISSUE #327: [Feature Request] Edit search query
Hello,
very nice project.

I've encountered an issue a few times where some books aren't found simply because their title includes a series part or number. A specific example is 'The Perfect Run'. The first part is found without any problems, but it fails to find the other parts (The Perfect Run II, The Perfect Run III) because the search includes 'II' and 'III' but not a number - 2 or 3. 

It would be very convenient if there was **an option to edit the search query before retrying the search.**

Pedro-Revez-Silva:
## Summary

Fixes #327 by making indexer search broaden automatically instead of adding a manual query input.

This keeps the search UX simple while covering the reported cases where provider metadata is more precise than indexer release naming. Shelfarr now tries additional generic indexer queries when structured or exact-title searches are empty or weak — `title author`, `author title`, subtitle-stripped titles, normalized titles, and roman numeral / arabic number variants such as `III` and `3` — and it no longer returns an empty result list when the broad pass retrieved plausible candidates.

## Why

Issue #327 asked for editable retry queries because releases such as `The Perfect Run II` / `The Perfect Run III` may exist under numeric naming in indexers, and a commenter reported success with `author title` ordering. Rather than shifting that work to the user, this PR improves the automatic search strategy and keeps broader matches ranked with lower confidence. The goal is that manual queries never feel necessary.

### Search broadening
- Add ordered generic indexer search attempts after the structured Prowlarr search: exact title, title + author, subtitle-stripped title (`Mistborn: The Final Empire` → `Mistborn`), author-first ordering, normalized title, and roman/arabic number variants.
- Apply confidence penalties to broader attempts and record the search attempt/query in `score_breakdown`.
- Normalize roman numerals during title scoring so `III` and `3` match strongly regardless of which query found the release.
- Standalone `I` is excluded from numeric variants (it is almost always the pronoun).

### Robustness
- Threshold checks now use the penalized score, so a broadened match cannot stop the search while being stored below the confidence threshold.
- A single failing generic query no longer discards results from other attempts; authentication errors still surface immediately.
- Fix indexer error constants in `Prowlarr`/`Jackett` singleton classes that raised `NameError` instead of typed errors on non-200 responses.

### Better category-less searching
- Releases that look like video content (resolution tags, `S01E05` markers, video codecs/sources) are rejected from category-less searches. Explicit book markers (EPUB, M4B, Audiobook, …) veto the detection so it stays high-precision.
- When nothing clears `min_match_confidence`, the best few low-confidence candidates are kept for manual review instead of returning an empty search.

### Documentation
- Document the search broadening behavior and `indexer_search_scope` in `docs/configuration.html`, and correct the `min_match_confidence` description.

## Validation

- `bundle exec rails test` (full suite, 1619 runs, 0 failures)
- `bundle exec rubocop` on all changed files
- Verified the issue scenarios end to end: `The Perfect Run II/III` generate `The Perfect Run 2/3` queries and score `title=100`; `author title` ordering is attempted when title-first queries are empty.

### DIFF
--- a/app/jobs/search_job.rb
+++ b/app/jobs/search_job.rb
@@ -3,6 +3,35 @@
 class SearchJob < ApplicationJob
   queue_as :default
 
+  SearchAttempt = Data.define(:name, :query, :score_penalty)
+
+  GENERIC_SEARCH_ATTEMPT_PENALTIES = {
+    exact_title: 0,
+    title_author: 5,
+    author_title: 8,
+    normalized_title: 10,
+    number_variant: 12
+  }.freeze
+
+  ROMAN_NUMERALS = {
+    "I" => "1",
+    "II" => "2",
+    "III" => "3",
+    "IV" => "4",
+    "V" => "5",
+    "VI" => "6",
+    "VII" => "7",
+    "VIII" => "8",
+    "IX" => "9",
+    "X" => "10",
+    "XI" => "11",
+    "XII" => "12",
+    "XIII" => "13",
+    "XIV" => "14",
+    "XV" => "15"
+  }.freeze
+  ARABIC_NUMERALS = ROMAN_NUMERALS.invert.freeze
+
   def perform(request_id)
     request = Request.find_by(id: request_id)
     return unless request
@@ -105,13 +134,13 @@ class SearchJob < ApplicationJob
 
   def search_indexer(request)
     if IndexerClient.provider == SearchResult::SOURCE_PROWLARR
-      results = search_prowlarr(request)
+      tagged_results = search_prowlarr(request)
     else
-      results = search_generic_indexer(request)
+      tagged_results = search_generic_indexer(request)
     end
 
-    results.map do |r|
-      { result: r, source: IndexerClient.provider }
+    tagged_results.map do |tagged|
+      tagged.merge(source: IndexerClient.provider)
     end
   end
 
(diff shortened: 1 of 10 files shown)


Pedro-Revez-Silva:
Reviewed the approach and pushed two follow-up commits.

The overall strategy (ordered broadening attempts with confidence penalties + roman numeral normalization in the scorer) is sound and correctly avoids pushing query editing onto the user. A few issues found and fixed:

**Fixes**

- **Typed indexer errors never actually raised** (`223d945`): `Prowlarr`/`Jackett` referenced `Error`/`AuthenticationError`/`ConnectionError` bare inside `class << self`, where Ruby constant lookup does not traverse `Base`'s constants — every non-200 response raised `NameError` instead of the typed error. Pre-existing on `main` (masked in tests by `.to_raise(...)` stubs), but this PR's error classification depends on those types, so fixed here with `Base::`-qualified constants.
- **Threshold checks now use the penalized score** (`03dea71`): `strong_indexer_match?` and `filter_broad_results` compared the *raw* score against `min_match_confidence`, while `save_results` stores score minus penalty. A number-variant match scoring 61 raw (49 after penalty) stopped all further broadening yet was persisted below the display/auto-select threshold. Both checks now score results exactly as they will be stored.
- **Per-attempt error isolation**: one failing generic query aborted the whole attempt loop and, on the Prowlarr supplement paths, discarded structured results already fetched. Individual attempt failures are now logged and skipped; `AuthenticationError` still propagates immediately (remaining attempts would fail identically), and a total failure with zero results still raises so retry scheduling keeps working.
- **Variant generation tweaks**: standalone `I` is excluded from roman→arabic variants ("I, Robot" no longer produces a "1, Robot" query; `1 → I` is kept since a lone digit is unambiguous), apostrophes are stripped instead of split ("Journey's End" → "Journeys End", not "Journey s End"), and the broad supplement guards against an empty attempt list.

**Tests added**

- broadening stops issuing queries once a strong match is found (no wasted indexer traffic)
- broadening continues when a match falls below the threshold only after its penalty
- results from later attempts survive an earlier attempt's 500
- 401 during generic attempts marks the request for attention
- no numeric variants generated for pronoun-`I` titles
- scorer: arabic book number matches roman release number (reverse direction of the existing test)

Full suite: 1610 runs, 0 failures. Rubocop clean.

PR #343, opened 2026-07-09
fix: safely split bundled audiobook imports

### LINKED ISSUE #337: [Feature Request] Split multi-file bundle torrents into per-book subfolders on import
When a torrent contains multiple audiobook files (e.g. a full series bundle with one .m4b per book), post-processing dumps all files into a single destination folder. This means Audiobookshelf (and other library managers) treats the entire series as one item, with hundreds of chapters jumbled together across all books — making it impossible to track progress per book.

## Expected behaviour

When the downloaded torrent directory contains multiple audio files, Shelfarr should create a separate subfolder for each file, applying the path template per file rather than per torrent.

For example, a "Throne of Glass" series torrent containing 7 .m4b files should produce:

```
/audiobooks/Sarah J. Maas/Throne of Glass (Throne of Glass, Book 1)/
    Throne of Glass (Throne of Glass, Book 1).m4b
/audiobooks/Sarah J. Maas/Crown of Midnight (Throne of Glass, Book 2)/
    Crown of Midnight (Throne of Glass, Book 2).m4b
...
```

instead of:

```
/audiobooks/Sarah J. Maas/Throne of Glass/
    Throne of Glass (Throne of Glass, Book 1).m4b
    Crown of Midnight (Throne of Glass, Book 2).m4b
    ...
```

## Suggested implementation

In `PostProcessingJob#import_files`, when handling a `directory_source`, detect if there are multiple primary audio files (`.m4b`, `.mp3`, `.flac`, `.aax`, etc.) and if so, create a subfolder per file rather than copying everything flat. This could be opt-in via a setting like **"Split bundle imports into per-book folders"**.

## Current workaround

Manually move each file into its own subfolder after download and re-scan the library.

Pedro-Revez-Silva:
## Summary
- add an opt-in setting for conservative multi-book audiobook bundle imports
- split only self-contained M4B/AAX siblings; MP3/FLAC/chaptered, multipart, nested, mixed, ambiguous, and AAXC releases keep the existing unsplit behavior
- prefer embedded metadata, then safe filename fallback, while preserving the originally requested book as Shelfarr's tracked `file_path`
- copy generic covers, exact-stem companions, and unrelated extras without data loss; extra bundle members remain filesystem/library-manager-only
- support flat path templates with per-title folders and pre-created tracked-book ZIPs

Closes #337

## Validation
- 1,794 Rails tests / 5,498 assertions
- focused bundle, post-processing, metadata, and live-update tests
- 30 seeded live-update test runs
- RuboCop on all touched files
- `bin/quality commit --staged`
- `bin/quality push`
- independent blocker review for multipart, move-import, companion, `file_path`, and deterministic Turbo refresh safety


hiSandog:
How does the conservative split handle two self-contained files whose embedded title metadata is identical but whose filenames identify different books? A useful regression case would verify that this ambiguity keeps the bundle unsplit, or otherwise documents which metadata source wins, so files cannot silently overwrite or merge into one destination.

Pedro-Revez-Silva:
> How does the conservative split handle two self-contained files whose embedded title metadata is identical but whose filenames identify different books? A useful regression case would verify that this ambiguity keeps the bundle unsplit, or otherwise documents which metadata source wins, so files cannot silently overwrite or merge into one destination.

Embedded metadata wins over the filename. If two files resolve to the same normalized embedded title,  `distinct_book_titles?` rejects the plan and the release remains unsplit. The existing “does not split duplicate embedded titles” regression test covers this ambiguity.

PR #350, opened 2026-07-11
fix: include path prefix in search results

### LINKED ISSUE #349: When using RAILS_RELATIVE_URL_ROOT some paths are not relative
When RAILS_RELATIVE_URL_ROOT is set to say `/books`, the main UI are correctly prefixed with /books.

However when the live search results are rendered, they're mainly missing the /books prefix, so the links do not work without adding an exception to the reverse proxy to pass these directly to the container.

It looks like when ActionController::Live returns control up the stack, it clears SCRIPT_NAME on the shared env Hash. Since Live's thread holds a reference to that same object, every subsequent chunk sees the corrupted value.

chartionic:
Capture request.script_name once at the very start of the action (before any commit can trigger the reset), then re-apply it via request.set_header before each subsequent render.

Fixes #349

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -45,6 +45,7 @@ class SearchController < ApplicationController
   end
 
   def stream_results
+    @relative_url_root = request.script_name
     @query = params[:q].to_s.strip
     @content_kind = normalized_content_kind(params[:content_kind])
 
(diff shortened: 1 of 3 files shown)


Pedro-Revez-Silva:
Thanks for investigating this — I reproduced the issue both on this PR’s original base and on the current `main`, and confirmed that these changes preserve the `/books` prefix in links rendered after the first streamed chunk.

Before merging, could you please:

- Rebase the branch onto the current `main`.
- Add a regression test that exercises the actual failure mode: mount the app under a prefix such as `/books` (for example with `Rack::URLMap`), allow the initial live-stream chunk to commit, then render a later provider-result chunk and assert that its internal links still begin with `/books`.
- Add a short comment explaining why `SCRIPT_NAME` has to be captured before the first stream write and reapplied for later renders. This lifecycle behavior is subtle and otherwise the mutation looks unnecessary.
- If practical, use the semantic `request.script_name=` API rather than setting the raw Rack header directly, and limit/restore the temporary mutation around `render_to_string` so the request environment is not left modified.

The core diagnosis and behavior of the fix look correct; the main missing piece is durable coverage for this streaming/Rack lifecycle edge case.

chartionic:
Thanks for the detailed feedback! I'll add the test tomorrow.

chartionic:
Sorry for the delay, I think I've addressed your feedback @Pedro-Revez-Silva .

Pedro-Revez-Silva (approved):
Thanks for tracking down the Live and SCRIPT_NAME lifecycle and for adding coverage for the mounted-path behavior.

Issue #338, opened 2026-07-05
Feat: Multiple ebook or Audiobook library


NathanIfinity:
Is there  way to monitor multiple audiobook / ebook library on book orbit? Right now my setup I  book orbit is I split my audio Library and ebook library in multiple library to better manage who Has access to what

Ex : I have multiple audiobook library (sci-fi, fantasy, litrpg, religion, self improvement, novel, etc..) and can only select one per type (audio or ebooks) in the settings

JoeJoeflyn:
Hi @NathanIfinity, thanks for raising this. Splitting libraries by genre/audience for access control is a reasonable BookOrbit setup.

From what I can see this would be more than a settings tweak — it'd need a backend change to store multiple library IDs per type, plus decisions on how delivery gets routed (e.g. requester picks, admin sets rules by genre), how duplicate detection scans across all configured libraries, and how sync loops through them. The frontend would follow
from that.

If you don't mind, one question that'd help before anyone builds it: does it actually break for you if delivery goes into one inbox library and BookOrbit handles the genre split and access control on its side? If so, what specifically breaks? That would help figure out whether the full multi-library model is needed or whether a lighter approachcovers your case.

Thanks again for the suggestion.

NathanIfinity:
It's more in regard to library scanning to know if the book is already in the library before the person request it, unless i don’t fully understand how it works if so I'm sorry for the unclear request. I thought the library you added in settings was the one that was scan to see if the requested book or audio book already existed.. 

In my case i don't actually use the full automation people request and when i find it (manually)I just delete the request and it shows up in  a scoped library where user see the new content !

NathanIfinity:
Wow thx for that change this is awesome!! 
Glad I was able to explain it properly lol

Pedro-Revez-Silva:
Implemented by #370. Multi-library BookOrbit inventories are now scanned for duplicate awareness while delivery remains assigned to the primary library.

PR #370, opened 2026-07-20
feat: scan multiple libraries per type for duplicate detection

JoeJoeflyn:
## Summary

References #338.

Adds support for scanning multiple libraries per type (audiobook / ebook / comics) for duplicate detection, without changing where downloads are delivered.

### What changed

- **3 new settings**: `audiobookshelf_*_scan_library_ids` — comma-separated additional library IDs to scan per type
- **Sync service**: `configured_library_ids` now merges the 3 delivery library IDs + 3 scan library IDs, splits on commas, dedupes
- **Path lookup**: `AudiobookshelfClient#find_item_by_path` does the same merge so path lookups also check scan libraries
- **Settings UI**: 3 new fields under Integrations → Library Platform. Multi-select dropdowns when a library platform is connected, text input fallback otherwise
- **Admin "In Library" pill**: amber badge on each request card in `/requests` when the requested book matches a synced `LibraryItem` from any configured library. Hover shows which library it's in

### Why

@NathanIfinity splits his BookOrbit libraries by genre (sci-fi, fantasy, litrpg, etc.) for access control and fulfills requests manually. Before this change, Shelfarr only scanned one library per type — books in his other genre libraries were invisible, so users kept requesting books he already had.

Now sync pulls items from all configured libraries, the matcher pools them, and both search results and the requests list surface "already in your library" hints across every library the admin owns. Delivery is unchanged — it still goes to the single primary library.


JoeJoeflyn:
Hi @Pedro-Revez-Silva, I noticed you force-pushed the PR. Could you let me know what changed compared to the previous version?

Pedro-Revez-Silva:
> Hi @Pedro-Revez-Silva, I noticed you force-pushed the PR. Could you let me know what changed compared to the previous version?

Rebased, nothing changed.

app/views/admin/settings/_form.html.erb
```diff
@@ -112,6 +112,14 @@
...
+                        <% elsif key.to_s.end_with?("_scan_library_ids") && @audiobookshelf_libraries.any? %>
+                          <% library_options = @audiobookshelf_libraries.map { |lib| ["#{lib.name} (#{lib.media_type})", lib.id] } %>
+                          <% selected_ids = data[:value].to_s.split(",").map(&:strip).reject(&:blank?) %>
+                          <%= form.select "settings[#{key}]", options_for_select(library_options, selected_ids), { multiple: true, include_blank: "Select libraries to scan for duplicates..." }, id: "settings_#{key}", class: "w-full rounded-md border border-gray-700 bg-gray-800 text-white focus:outline-none focus:ring-2 focus:ring-blue-500 focus:border-blue-500 px-3 py-2 min-h-[80px]", data: { action: "change->settings-form#autoSave" } %>
```
Pedro-Revez-Silva: This does not render a multi-select. The third hash passed to `form.select` contains select options; `multiple` must be in the fourth `html_options` hash. As written, the generated field lacks the `multiple` attribute and array-style name, so it can submit only one library ID and can collapse an existing list on save. Please move `multiple: true` into the HTML options and add a rendered-form round-trip test.

app/controllers/requests_controller.rb
```diff
@@ -36,6 +36,16 @@ def index
     base_requests = Current.user.admin? ? Request : Request.for_user(Current.user)
     @attention_count = base_requests.needs_attention.count
     @active_count = base_requests.active.where(attention_needed: false).count
+
+    # Preload library matches for admin (used for "In Library" pill on each card)
+    @library_matches_by_book = if Current.user&.admin? && LibraryItem.available_for_matching.exists?
+      matcher = AudiobookshelfLibraryMatcherService.new
+      @requests.each_with_object({}) do |req, hash|
+        hash[req.book_id] = matcher.matches_for(title: req.book.title, author: req.book.author, limit: 1)
```
Pedro-Revez-Silva: This makes the unpaginated request page perform full fuzzy matching for every request. `matches_for` iterates and scores the entire cached inventory each time, so this is `O(requests * library_items)` and adding several scan libraries can make `/requests` time out. Please paginate or batch/deduplicate the visible books and avoid rescanning the full inventory per request.

Pedro-Revez-Silva: The scan settings are per media type, but this passes only title and author to a matcher backed by every active-platform `LibraryItem`. An ebook-library copy can therefore mark an audiobook request for the same title as `In Library`. Please scope candidates to the delivery and scan IDs for `req.book.book_type`, or persist and filter the item media type.

app/services/audiobookshelf_library_sync_service.rb
```diff
@@ -52,8 +52,11 @@ def configured_library_ids
       configured_ids = [
         SettingsService.get(:audiobookshelf_audiobook_library_id),
         SettingsService.get(:audiobookshelf_ebook_library_id),
-        SettingsService.get(:audiobookshelf_comicbook_library_id)
-      ].filter_map(&:presence).uniq
+        SettingsService.get(:audiobookshelf_comicbook_library_id),
+        SettingsService.get(:audiobookshelf_audiobook_scan_library_ids),
+        SettingsService.get(:audiobookshelf_ebook_scan_library_ids),
+        SettingsService.get(:audiobookshelf_comicbook_scan_library_ids)
+      ].flat_map { |id| id.to_s.split(",").map(&:strip) }.filter_map(&:presence).uniq
```
Pedro-Revez-Silva: Removing a scan-library ID never removes or excludes its cached rows. The next sync simply stops visiting that library, while stale deletion only occurs inside libraries that are visited; `LibraryItem.available_for_matching` continues returning the old rows indefinitely. Please prune active-platform rows outside the current configured ID set, or constrain matching to that set.

Pedro-Revez-Silva: These additional IDs now enter destructive reconciliation, but the Audiobookshelf and BookOrbit `library_items` methods return an empty or partial array on a non-200 page. `sync_library_items` then deletes every unseen cached row and this loop counts the library as successfully synced. A transient API failure can erase duplicate-detection data. Please raise on non-success responses and prune only after a complete fetch.

app/views/requests/index.html.erb
```diff
@@ -109,6 +109,14 @@
                     <span class="inline-flex items-center px-2 py-1 rounded text-xs font-medium <%= request_status_color(request.status) %>">
                       <%= request.status.humanize %>
                     </span>
+                    <% if @library_matches_by_book&.dig(request.book_id)&.any? %>
```
Pedro-Revez-Silva, line 112: `matches_for` includes `:possible` matches at scores from 85 through 99, but this condition presents every match as definitively `In Library`. The search UI deliberately calls these related titles and shows `Possible match`; this badge can encourage an admin to cancel a valid request. Please reserve `In Library` for likely matches or render possible matches with non-definitive wording.

Pedro-Revez-Silva (changes requested):
Requesting changes for the broken multi-library selector, request-page matching correctness and scalability, and unsafe cache reconciliation. Inline comments include concrete failure scenarios and suggested fix directions.

JoeJoeflyn:
Pedro, thanks for the review — all six points were fair, fixes pushed:

  1. Multi-select rendering fixed
  2. API failures no longer wipe the cache
  3. Stale rows pruned when scan libraries are removed
  4. Matching scoped to the right book type
  5. Deduped and SQL-filtered to cut matching work
  6. "In Library" reserved for exact matches, fuzzy ones say "Possibly
  in Library"

The data-loss one on #5 was a real miss — good catch.

app/services/audiobookshelf_library_sync_service.rb
```diff
@@ -36,6 +36,9 @@ def sync!
       end
     end
 
+    # Prune cached rows for libraries that are no longer configured.
+    LibraryItem.for_platform(library_platform).where.not(library_id: library_ids).delete_all
```
Pedro-Revez-Silva: This platform-wide prune is unsafe when sync jobs overlap. The queue runs three worker threads and this job has no concurrency limit: an older run can snapshot `old-lib`, a newer run can sync `new-lib`, and then the older run resumes and deletes every `new-lib` row. Please serialize library syncs and revalidate the active platform/configured-ID snapshot before pruning, or protect rows written after this run started.

app/controllers/requests_controller.rb
```diff
@@ -36,7 +36,47 @@ def index
...
+    else
+      {}
+    end
+  end
+
+  def library_ids_for_book_type(book_type)
+    delivery_key, scan_key = case book_type.to_s
+    when "audiobook"  then [ :audiobookshelf_audiobook_library_id, :audiobookshelf_audiobook_scan_library_ids ]
+    when "ebook"      then [ :audiobookshelf_ebook_library_id, :audiobookshelf_ebook_scan_library_ids ]
+    when "comicbook"  then [ :audiobookshelf_comicbook_library_id, :audiobookshelf_comicbook_scan_library_ids ]
```
Pedro-Revez-Silva: This helper does not use the existing effective library-resolution rules. Comics are deliberately delivered to the ebook library when no comic library is configured (`SettingsService.library_id_for_book`), but this returns only comic-specific IDs. Also, when every explicit ID is blank, sync auto-discovers libraries while this helper returns `[]`, suppressing every badge. Please derive this scope from shared fallback logic and preserve all-library matching when synchronization used discovery.

app/views/admin/settings/_form.html.erb
```diff
@@ -112,6 +112,14 @@
...
+                        <% elsif key.to_s.end_with?("_scan_library_ids") && @audiobookshelf_libraries.any? %>
+                          <% library_options = @audiobookshelf_libraries.map { |lib| ["#{lib.name} (#{lib.media_type})", lib.id] } %>
+                          <% selected_ids = data[:value].to_s.split(",").map(&:strip).reject(&:blank?) %>
+                          <%= form.select "settings[#{key}]", options_for_select(library_options, selected_ids), { include_blank: "Select libraries to scan for duplicates..." }, { id: "settings_#{key}", multiple: true, class: "w-full rounded-md border border-gray-700 bg-gray-800 text-white focus:outline-none focus:ring-2 focus:ring-blue-500 focus:border-blue-500 px-3 py-2 min-h-[80px]", data: { action: "change->settings-form#autoSave" } } %>
```
Pedro-Revez-Silva, line 130: Options come only from the current API response. If a persisted selected ID is omitted from a partial response, it disappears from the field; the next full-form autosave submits only the hidden blank or remaining visible IDs, clears the missing setting, and the new sync prune deletes its cache. Please merge unavailable persisted IDs into the options as selected entries and add a full autosave round-trip test.

app/services/audiobookshelf_client.rb
```diff
@@ -58,7 +58,7 @@ def library_items(id, page_size: 500)
 
       loop do
         response = request { connection.get("/api/libraries/#{id}/items", { limit: page_size, page: page }) }
-        break unless response.status == 200
+        raise Error, "Audiobookshelf library #{id} returned status #{response.status}" unless response.status == 200
```
Pedro-Revez-Silva, line 64: Raising for transient failures correctly preserves the cache, but treating every non-200 identically means a first-page `404` or `410` for a permanently removed library also preserves stale matches indefinitely. Please distinguish authoritative first-page absence from authentication, rate-limit, 5xx, and later-page failures. The same issue exists in `BookOrbitClient#library_items`.

app/controllers/requests_controller.rb
```diff
@@ -36,7 +36,47 @@ def index
     base_requests = Current.user.admin? ? Request : Request.for_user(Current.user)
     @attention_count = base_requests.needs_attention.count
     @active_count = base_requests.active.where(attention_needed: false).count
+
+    # Pagination to limit fuzzy matching overhead
+    @requests_page = params[:page].to_i.clamp(1, 1_000_000)
+    @requests_per_page = 25
+    @requests_total_count = @requests.count
+    @requests_total_pages = [ (@requests_total_count.to_f / @requests_per_page).ceil, 1 ].max
+    @requests = @requests.limit(@requests_per_page).offset((@requests_page - 1) * @requests_per_page)
```
Pedro-Revez-Silva, line 46: The requested page is capped globally but never clamped to `@requests_total_pages`. A stale bookmark or `/requests?page=999` produces an empty relation, renders `No requests yet`, and provides no recovery link because pagination is inside the non-empty branch. Please clamp or redirect to the last valid page after calculating the total and before applying the offset.

Pedro-Revez-Silva (changes requested):
The follow-up passes all CI checks, but the focused adversarial review found regressions around concurrent cache reconciliation, effective library fallback resolution, settings round-trips, authoritative API absence, and pagination boundaries. Inline comments include concrete failure scenarios and fix directions.

JoeJoeflyn:
Thanks for the feedback! I've addressed all the review points:

  1. Sync Concurrency: Serialized runs using a Mutex, revalidated configured IDs
  before pruning, and limited deletes to items synced before this run started
  (synced_at < now).
  2. Library Fallback: Integrated fallback logic in the requests controller to
  route comicbooks to ebooks when comic libraries are empty, and return nil
  (preserve all-library matching) when auto-discovery is active.
  3. Settings Round-Trips: Persisted and merged unavailable library IDs as
  (Unavailable) options in form selects so they are not cleared during autosave.
  4. Authoritative Absence: Clients now return [] on page 0 404/410 errors to
  allow sync to prune the removed library's cache.
  5. Pagination: Clamped page param to @requests_total_pages to avoid empty
  relations on stale page bookmarks.

  Added unit and integration tests covering each scenario. All checks are green!

app/services/audiobookshelf_library_sync_service.rb
```diff
@@ -7,7 +7,17 @@ def success?
     end
   end
 
+  SYNC_MUTEX = Mutex.new
```
Pedro-Revez-Silva: This mutex only serializes threads inside one Ruby process. `config/queue.yml` explicitly permits multiple Solid Queue processes through `JOB_CONCURRENCY`; each process owns a different mutex, so an older run can still reconcile after a newer run and delete newer rows in `sync_library_items`. Please use durable job-level `limits_concurrency to: 1`, a database-backed lock, or a run-generation guard spanning all processes.

app/controllers/requests_controller.rb
```diff
@@ -36,7 +36,67 @@ def index
...
+
+    delivery_key, scan_key = case book_type.to_s
+    when "audiobook"
+      [ :audiobookshelf_audiobook_library_id, :audiobookshelf_audiobook_scan_library_ids ]
+    when "comicbook"
+      if SettingsService.get(:audiobookshelf_comicbook_library_id).present?
+        [ :audiobookshelf_comicbook_library_id, :audiobookshelf_comicbook_scan_library_ids ]
+      else
+        [ :audiobookshelf_ebook_library_id, :audiobookshelf_ebook_scan_library_ids ]
+      end
```
Pedro-Revez-Silva: When comic delivery falls back to the ebook library, this also replaces the comic scan setting with the ebook scan setting. A configured `audiobookshelf_comicbook_scan_library_ids` value is therefore ignored, while unrelated ebook scan libraries are included. Only the delivery ID should fall back to ebook; comic requests should always retain the comic-specific scan IDs.

app/services/audiobookshelf_library_sync_service.rb
```diff
@@ -36,6 +46,17 @@ def sync!
       end
     end
 
+    # Prune cached rows for libraries that are no longer configured.
+    # Revalidate active platform and configured IDs snapshot, and protect rows written after this run started.
+    remove_instance_variable(:@configured_library_ids) if defined?(@configured_library_ids)
+    current_library_ids = configured_library_ids
```
Pedro-Revez-Silva: In auto-discovery mode this clears the discovered-ID snapshot and performs a second external libraries request outside the per-library rescue. If that request fails, `sync!` raises after already mutating the cache; if it returns a transiently partial list, the prune can remove retained rows for a library whose item sync failed. Reuse the initial authoritative discovery result, or catch revalidation failures and skip destructive pruning for that run.

app/services/audiobookshelf_library_sync_service.rb
```diff
@@ -36,6 +46,17 @@ def sync!
...
+    current_platform = LibraryPlatformClient.active_platform
+
+    LibraryItem.for_platform(current_platform)
+               .where.not(library_id: current_library_ids)
+               .where("synced_at < ?", now)
```
Pedro-Revez-Silva: Rows written by this same run use `synced_at == now`, so if configuration changes during the run this strict `< now` predicate preserves the now-unconfigured rows. Rows with a null timestamp also never satisfy the SQL comparison. After durable serialization, use a run generation or include `synced_at <= now` and null stale rows in the cleanup.

test/controllers/requests_controller_test.rb
```diff
@@ -2119,4 +2184,61 @@ def create_store_offer
       quoted_at: Time.current
     )
   end
+
+  test "index preloads library matches utilizing comic fallback and auto-discovery" do
```
Pedro-Revez-Silva: This test is declared after the class-level `private` section. Minitest discovers tests through `public_instance_methods`, so this fallback/auto-discovery test is not executed. Move it above `private` so the intended regression coverage actually runs.

Pedro-Revez-Silva (changes requested):
The latest commit fully addresses unavailable-ID preservation, authoritative page-zero 404/410 handling, and pagination bounds. However, the concurrency fix remains incomplete across worker processes, and the commit introduces regressions in comic scan scoping, discovery revalidation, timestamp pruning, and test discovery. Inline comments provide reproduction scenarios and fix directions.

JoeJoeflyn:
All regressions fixed: serialized sync concurrency at job level, fixed comic scan scoping fallback, removed redundant discovery API calls, corrected prune query, and moved tests out of private. All tests pass!

app/jobs/audiobookshelf_library_sync_job.rb
```diff
@@ -2,6 +2,9 @@
 
 class AudiobookshelfLibrarySyncJob < ApplicationJob
   queue_as :default
+  limits_concurrency to: 1,
+    key: "audiobookshelf-library-sync",
+    duration: 10.minutes
```
Pedro-Revez-Silva, line 7: Solid Queue treats `duration` as the maximum concurrency guarantee: after ten minutes a blocked job may acquire the semaphore even while the first job is still running. This sync has no ten-minute upper bound because it paginates every configured library and persists items individually. An older run can therefore overlap a newer run and later delete newer rows. Please use a durable run-generation/database guard, or otherwise retain correctness after lease expiry rather than relying solely on this fixed duration.

app/services/audiobookshelf_library_sync_service.rb
```diff
@@ -36,6 +36,12 @@ def sync!
       end
     end
 
+    # Prune cached rows for libraries that are no longer configured.
+    LibraryItem.for_platform(library_platform)
+               .where.not(library_id: library_ids)
```
Pedro-Revez-Silva: `library_ids` was captured before the API work, but settings can change while this job runs. If library B is configured during a sync that started with only A, this old snapshot still treats B as unconfigured and deletes existing B rows whose timestamp is at or before `now`; the queued follow-up may be blocked or fail. Re-read the local explicit platform/ID settings before pruning and skip destructive cleanup if they changed, while reusing the initial discovery result to avoid another external request.

Pedro-Revez-Silva (changes requested):
The latest commit fully addresses comic scan fallback, duplicate discovery, inclusive/null pruning, and test visibility. Two synchronization risks remain: the concurrency guarantee expires after ten minutes, and destructive pruning uses a stale initial settings snapshot. Inline comments describe concrete races and fix directions.

JoeJoeflyn:
thanks for feedback Pedro. Addressed both concurrency & settings race conditions in the latest push:

- **Concurrent Run Overlap Safety:** `sync_library_items` and item pruning now guard against `synced_at > run_start`. If a newer sync run finishes or updates items concurrently, an older run will not overwrite their metadata or delete their synced rows.
- **Mid-Run Settings Revalidation:** Explicit platform/library settings are re-evaluated right before pruning. If settings were changed mid-run, destructive pruning of unconfigured libraries is skipped while retaining the initial discovery snapshot without extra API calls.

Pedro-Revez-Silva (approved):

PR #377, opened 2026-07-21
feat: add PostgreSQL support alongside SQLite

agreenbhm:
This change introduces PostgreSQL support to the application while preserving SQLite as the default database.

Key changes:
- Adds the `pg` gem to the `Gemfile` and locks dependencies correctly.
- Updates `config/database.yml` to switch database adapter dynamically based on the `DB_ADAPTER` environment variable.
- Refactors custom SQLite functions in `app/controllers/library_controller.rb` to use conditionally correct SQL syntax (`REGEXP_REPLACE`, `STRING_AGG`, `json_array_elements_text`) for PostgreSQL when it's the active connection adapter.
- Adjusts database tests in `test/migrations` to clean up and cascade dropped tables natively when running in PostgreSQL instead of relying on the in-memory isolated `sqlite3` driver.
- Adds `libpq-dev` and `postgresql-client` system dependencies to `Dockerfile` to enable building `pg` in production.
- Resolved various Rubocop styling infractions across the repository.


Pedro-Revez-Silva:
Thanks for taking the time to put this together. PostgreSQL support needs broader design work across configuration, migrations, runtime queries, testing, deployment, and data migration, so we cannot merge this implementation as submitted. Please open a feature request in GitHub Issues describing the use case and deployment requirements first. That will let us agree on the scope and approach before another implementation is started.

agreenbhm:
I disagree, this isn't enterprise software and in testing my PR outperforms SQLite significantly. Migration isn't something that is generally officially supported by arr apps, though there are usually instructions on how to use something like pgloader with no promises. Obviously it's your call since this is your project, but I think the pros of merging this now outweigh the cons. Just my 2 cents.

Pedro-Revez-Silva:
Your PR had a bunch of problems that makes it easier to just implement it myself instead of spending my time and tokens on back and forth. 

I didn't say no to adding psql. The performance is only noticeable at numbers users of this app are unlikely to get to, and it doesn't allow me to package the DB along with the app, making the installation harder. 

Anyway, I will look into it.

PR #404, opened 2026-07-28
fix: send the author to Prowlarr indexers that only accept free text

coffbr01:
**If you'd rather implement this yourself, please do — everything needed to reproduce, diagnose and re-derive the fix is written out below, and the patch can be discarded without losing anything.**

## The problem

For anyone whose indexers are free-text only — which is most torrent trackers — **the structured book search retrieves nothing at all**, and the author never reaches any indexer.

`IndexerClients::Prowlarr.build_query` builds a correct structured query and `search` issues it as `type=book`:

```ruby
parts << "{title:#{sanitized_title}}"   if sanitized_title.present?
parts << "{author:#{sanitized_author}}" if sanitized_author.present?
```

Prowlarr parses those tokens into `BookSearchCriteria.Title` / `.Author` (`NewznabRequest.QueryToParams()`), and then, in `src/NzbDrone.Core/Indexers/HttpIndexerBase.cs`:

```csharp
public override Task<IndexerPageableQueryResult> Fetch(BookSearchCriteria searchCriteria)
{
    ...
    if ((searchCriteria.Title.IsNotNullOrWhiteSpace() && !caps.BookSearchTitleAvailable) ||
        (searchCriteria.Author.IsNotNullOrWhiteSpace() && !caps.BookSearchAuthorAvailable) ||
        (searchCriteria.Publisher.IsNotNullOrWhiteSpace() && !caps.BookSearchPublisherAvailable) ||
        (searchCriteria.Genre.IsNotNullOrWhiteSpace() && !caps.BookSearchGenreAvailable) ||
        (searchCriteria.Year.HasValue && !caps.BookSearchYearAvailable))
    {
        _logger.Debug("Book search skipped due to unsupported capabilities used: {0}", Definition.Name);
```

It **skips the indexer entirely** rather than falling back to the free text. On a `bookSearchParams: ['q']` indexer the search returns zero, in under 10 ms, having made no HTTP request and left no row in Prowlarr's history.

## Reproduce it in one command

Any Prowlarr with a `q`-only book indexer (check `GET /api/v1/indexer` → `capabilities.bookSearchParams`). This is the exact query Shelfarr sends:

```bash
# what Shelfarr sends today -> 0 results, indexer never contacted
curl -s -G -H "X-Api-Key: $KEY" http://prowlarr:9696/api/v1/search \
  --data-urlencode 'query={title:Frankenstein} {author:Mary Shelley}' \
  --data-urlencode 'type=book' | jq length

# same indexer, free text -> the real results
curl -s -G -H "X-Api-Key: $KEY" http://prowlarr:9696/api/v1/search \
  --data-urlencode 'query=Frankenstein Mary Shelley' \
  --data-urlencode 'type=book' | jq length
```

Measured against a live Prowlarr 2.4.0.5397 / Shelfarr 0.39.2, one `q`-only indexer:

| query | type | result |
|---|---|---|
| `{title:Frankenstein} {author:Mary Shelley}` | book | **0 results, <10 ms, indexer never contacted** |
| the same tokens **+** free text appended | book | **0** — the token's mere presence triggers the skip |
| `Frankenstein Mary Shelley` | book | **20 results** |
| `Frankenstein Mary Shelley` | search | 20 results |

★ **The diagnostic tell is that a `type=book` search leaves NO ROW in Prowlarr's history** (`GET /api/v1/history?sortKey=date&sortDirection=descending`). A skipped indexer and a genuinely-empty search are indistinguishable from Shelfarr's side, since Shelfarr doesn't log its outbound query — Prowlarr's history is the only honest witness. That is also how to verify any fix: **pass = the recorded query contains the author.**

## Why it stays invisible

Because the structured search comes back empty, `search_prowlarr` falls through to `search_generic_indexer_attempts`, whose first attempt is `exact_title` — **title only**. `strong_indexer_match?` then breaks the attempt loop as soon as anything clears `min_match_confidence`, so the `title_author` and `author_title` attempts a few lines below **never run**.

So **the author drives ranking but never retrieval.** The scorer is author-aware and does its job — it just can only rank what the query dragged in. What a title-only query actually drags in, across four indexers in the ebook categories:

```
"Frankenstein"  ->  126 candidates, 12 of which name Shelley

  Frankenstein by Mary Shelley EPUB
  Cleopatra and Frankenstein by Coco Mellors EPUB
  Frankenstein in Baghdad by Ahmed Saadawi EPUB
  Criminal Macabre - The Eyes of Frankenstein 04 (of 04) (2013)
  Frankenstein and Philosophy by Nicolas Michaud
  Frankenstein Agent of S.H.A.D.E. v1 (000 - 016) (2011 - 2013)
  Tales of Terror (Dracula, Frankenstein, The Legend of Sleepy Hollow)
  Dean Koontz's Frankenstein - Prodigal Son v2 (01 - 05) (2010)
  Young Frankenstein_ The Story of the Making of the Film by Mel Brooks
  Young.Frankenstein.1974.480p.BRRip.XviD-SHiRK
  ...
```

Nine in ten candidates are a different book, a comic, a making-of, or a 1974 film. A good scorer on a pool like that produces a confident-looking wrong answer: in the request that surfaced this for me, the top-ranked result sat at **confidence 96** and was a **1-seeder** public-tracker torrent, ranked above four unrelated books with similar titles, while a several-hundred-seeder copy of the right book on another indexer was never asked for.

It applies to every request and every `q`-only indexer, always. Merely survivable for a distinctive title; severe for a one-word one.

## If you'd rather write it yourself

The whole fix fits in `app/services/indexer_clients/prowlarr.rb`. The shape I'd suggest, and the decisions any implementation has to make:

1. **Capabilities are already one call away.** `GET /api/v1/indexer` returns `capabilities.bookSearchParams` per indexer, and `Prowlarr.indexers` already fetches it for tag filtering — so a capability-aware query costs no new endpoint, and no extra request at all when `prowlarr_tags` is set.
2. **Partition, don't pick one shape.** A mixed setup (a Newznab book indexer alongside torrent trackers) needs the structured query for the former and free text for the latter, i.e. one `api/v1/search` call per group with explicit `indexerIds`. Choosing a single shape for everyone regresses one half or the other.
3. **Only emit a token you can prove is supported.** An indexer qualifies for the structured query only if it advertises *every* field being sent — sending `{title:}` to something that supports `author` but not `title` re-triggers the same skip.
4. **Keep `type=book`.** Free text works fine under `type=book` (see the table above), and it preserves category mapping. Downgrading to `type=search` would lose that.
5. **Degrade toward free text, never toward tokens.** If the indexer list can't be read, or an indexer reports no capabilities, free text is the shape no indexer can reject.
6. **Appending free text to the tokens does not work** — worth knowing before trying it, since it looks like the cheapest fix. The token's presence triggers the skip regardless of what else is in `q` (row 2 of the table).

An alternative worth naming and rejecting: always sending free text and dropping the structured query entirely. It fixes the reported symptom in one line, but throws away real precision on Newznab/usenet indexers that *do* advertise `author`/`title`.

## What this PR does

Exactly the above:

- advertises `author`/`title` → the structured `{title:}{author:}` query, unchanged
- otherwise → `"<title> <author>"` as free text, still `type=book`
- indexer list unreadable → free text

Common all-`q`-only and all-structured setups stay a single request; only genuinely mixed setups issue two.

## Verification

Same environment, same book, through the patched client: **31 candidates, 29 of which name Shelley**, topped by a 319-seeder EPUB — against **0** before, and against 126-candidates-12-of-which-are-the-book from the title-only fallback that was doing all the real work. Prowlarr history for that run:

```
idx=12 | query='Frankenstein Mary Shelley' | type=book | results=40
idx=10 | query='Frankenstein Mary Shelley' | type=book | results=20
idx=11 | query='Frankenstein Mary Shelley' | type=book | results=5
idx=7  | query='Frankenstein Mary Shelley' | type=book | results=4
```

Audiobooks share this path, so they were checked separately (categories 3030):

| query | results | actually the right book |
|---|---|---|
| tokens (today) | **0** | 0 |
| free text (this PR) | 20 | **20** |
| title-only fallback | 20 | 8 |

The new primary query is narrower than a bare title, so a release whose name omits the author can be missed by it — but the title-only generic attempts still fire whenever no strong match is found, so that safety net is unchanged.

Running in production on my own stack since 2026-07-27, verified end to end in the app as well: a newly filed request now puts `query='<title> <author>'` `type=book` on the wire to all four indexers, where the book search previously produced no history row at all.

## Tests

- `test/services/prowlarr_client_test.rb`: free-text query for a `q`-only indexer; the split across a mixed indexer set; the fallback when capabilities are unavailable. The two existing structured-query tests now stub an indexer advertising `author`/`title`.
- `test/jobs/search_job_test.rb`: a job-level regression test asserting the author reaches a `q`-only indexer with no `{title:}`/`{author:}` tokens, plus a default capability stub in `setup` (the client now reads `/api/v1/indexer` on the book path, which most existing tests didn't stub).

Full suite green, `rubocop` clean. One pre-existing unrelated failure (`FileCopyServiceTest#test_remove_source_file_retains_an_in-place_mutation_after_snapshot`) fails identically on a clean `main`.

## Unrelated observation

`generic_indexer_query` looks like dead code on `main` — defined at `app/jobs/search_job.rb:343`, never called. Not touched here.

### DIFF
--- a/app/services/indexer_clients/prowlarr.rb
+++ b/app/services/indexer_clients/prowlarr.rb
@@ -19,8 +19,12 @@ module IndexerClients
           return items.map { |item| parse_result(item) }
         end
 
-        scoped = scoped_indexers
-        plans = book_search_plans(query, title: title, author: author, indexers: scoped)
+        # Resolve the configured tag scope once and carry it through. Reading it
+        # a second time while planning lets a transient /api/v1/tag failure read
+        # as "no tags configured" and widen the search past the user's scope.
+        tags = indexer_filter_tags
+        scoped = scoped_indexers(tags: tags)
+        plans = book_search_plans(query, title: title, author: author, tags: tags, indexers: scoped)
         items = execute_book_search_plans(plans, categories: cats, limit: limit)
 
         deduplicate_by_guid(items, indexer_priorities(scoped)).map { |item| parse_result(item) }
@@ -292,7 +296,7 @@ module IndexerClients
       # accept and give each group the richest query it supports. A token is
       # only ever emitted when the indexer is known to support it; everything
       # else falls back to free text, which every indexer accepts.
-      def book_search_plans(query, title:, author:, indexers: scoped_indexers)
+      def book_search_plans(query, title:, author:, tags: indexer_filter_tags, indexers: scoped_indexers(tags: tags))
         free_text_query = build_free_text_query(query, title: title, author: author)
 
         # Indexer list unavailable, or nothing matched the configured tags: keep
@@ -317,8 +321,9 @@ module IndexerClients
         # One plan covers every indexer Prowlarr would have searched anyway, so
         # drop the selection and leave the choice to Prowlarr exactly as before
         # this split existed. An indexer blocked since the list was read then
-        # costs nothing.
-        plans.first[:indexer_ids] = nil if plans.one? && indexer_filter_tags.empty?
+        # costs nothing. Only safe without a tag scope: with one, "every indexer
+        # Prowlarr would search" is the whole enabled set, not the tagged subset.
+        plans.first[:indexer_ids] = nil if plans.one? && tags.empty?
 
         plans
       end
--- a/app/services/indexer_clients/prowlarr.rb
+++ b/app/services/indexer_clients/prowlarr.rb
@@ -8,22 +8,14 @@ module IndexerClients
       def search(query, categories: nil, book_type: nil, limit: 100, title: nil, author: nil)
         ensure_configured!
 
-        params = {
-          query: build_query(query, title: title, author: author),
-          type: search_type_for(title: title, author: author),
-          limit: limit
-        }
-
         cats = categories || categories_for_type(book_type)
-        params[:categories] = Array(cats) if cats.present?
 
-        indexer_ids = filtered_indexer_ids
-        params[:indexerIds] = indexer_ids if indexer_ids.present?
+        if search_type_for(title: title, author: author) == "search"
+          return execute_search(query, type: "search", categories: cats, limit: limit, indexer_ids: filtered_indexer_ids)
+        end
 
-        response = request { connection.get("api/v1/search", params) }
-
-        handle_response(response) do |data|
-          Array(data).map { |item| parse_result(item) }
+        book_search_plans(query, title: title, author: author).flat_map do |plan|
+          execute_search(plan[:query], type: "book", categories: cats, limit: limit, indexer_ids: plan[:indexer_ids])
         end
       end
 
@@ -34,16 +26,25 @@ module IndexerClients
         handle_response(response) { |data| Array(data) }
       end
 
+      # Indexers Prowlarr would search on our behalf: every configured indexer,
+      # or only the tagged subset when prowlarr_tags is set. Returns nil when
+      # Prowlarr's indexer list cannot be read.
+      def scoped_indexers(tags: indexer_filter_tags)
+        normalized_tags = tags.map(&:to_s).map(&:downcase)
+        all = indexers
+        return all if normalized_tags.empty?
+
+        all.select { |indexer| (normalized_indexer_tags(indexer["tags"]) & normalized_tags).any? }
+      rescue IndexerClients::Base::Error => e
+        Rails.logger.warn "[IndexerClients::Prowlarr] Failed to fetch indexers: #{e.message}"
+        nil
+      end
+
       def filtered_indexer_ids
-        tags = indexer_filter_tags.map(&:to_s).map(&:downcase)
+        tags = indexer_filter_tags
         return nil if tags.empty?
 
-        indexers
-          .select { |indexer| (normalized_indexer_tags(indexer["tags"]) & tags).any? }
-          .map { |indexer| indexer["id"] }
-      rescue IndexerClients::Base::Error => e
-        Rails.logger.warn "[IndexerClients::Prowlarr] Failed to fetch indexers for tag filtering: #{e.message}"
-        nil
+        scoped_indexers(tags: tags)&.map { |indexer| indexer["id"] }
       end
 
       def configured_tags
@@ -202,6 +203,71 @@ module IndexerClients
         nil
       end
 
+      def execute_search(query, type:, categories:, limit:, indexer_ids:)
+        params = { query: query, type: type, limit: limit }
+        params[:categories] = Array(categories) if categories.present?
+        params[:indexerIds] = indexer_ids if indexer_ids.present?
+
+        response = request { connection.get("api/v1/search", params) }
+
+        handle_response(response) do |data|
+          Array(data).map { |item| parse_result(item) }
+        end
+      end
+
+      # Prowlarr does not degrade a book search gracefully: when the query
+      # carries a {title:} or {author:} token that an indexer does not
+      # advertise in its bookSearchParams, that indexer is skipped outright
+      # ("Book search skipped due to unsupported capabilities used") and
+      # contributes nothing, rather than being queried on the free text.
+      # Most torrent trackers advertise `q` only, so a structured query
+      # retrieves literally zero results there while looking healthy.
+      #
+      # Group the indexers we are about to search by what they can actually
+      # accept and give each group the richest query it supports. A token is
+      # only ever emitted when the indexer is known to support it; everything
+      # else falls back to free text, which every indexer accepts.
+      def book_search_plans(query, title:, author:)
+        structured, free_text = Array(scoped_indexers).partition do |indexer|
+          supports_structured_book_search?(indexer, title: title, author: author)
+        end
+
+        plans = []
+        plans << search_plan(build_query(query, title: title, author: author), structured) if structured.any?
+        plans << search_plan(build_free_text_query(query, title: title, author: author), free_text) if free_text.any?
+
+        return plans if plans.any?
+
+        # Indexer list unavailable (or nothing matched the configured tags):
+        # search everything with the query shape no indexer can reject.
+        [ search_plan(build_free_text_query(query, title: title, author: author), []) ]
+      end
+
+      def search_plan(query, indexers)
+        { query: query, indexer_ids: indexers.map { |indexer| indexer["id"] }.presence }
+      end
+
+      def supports_structured_book_search?(indexer, title:, author:)
+        params = book_search_params(indexer)
+        return false if params.nil?
+        return false if sanitized_book_value(title).present? && !params.include?("title")
+        return false if sanitized_book_value(author).present? && !params.include?("author")
+
+        true
+      end
+
+      def book_search_params(indexer)
+        return nil unless indexer.is_a?(Hash)
+
+        capabilities = indexer["capabilities"]
+        return nil unless capabilities.is_a?(Hash)
+
+        params = capabilities["bookSearchParams"]
+        return nil unless params.is_a?(Array)
+
+        params.map { |param| param.to_s.downcase }
+      end
+
       def search_type_for(title:, author:)
         sanitized_book_value(title).present? || sanitized_book_value(author).present? ? "book" : "search"
       end
@@ -217,6 +283,13 @@ module IndexerClients
         parts << "{author:#{sanitized_author}}" if sanitized_author.present?
         parts << query if query.present?
         parts.join(" ")
+      end
+
+      def build_free_text_query(query, title:, author:)
+        [ sanitized_book_value(title), sanitized_book_value(author), query ]
+          .compact_blank
+          .join(" ")
+          .squish
       end
 
       def sanitized_book_value(value)
--- a/app/services/indexer_clients/prowlarr.rb
+++ b/app/services/indexer_clients/prowlarr.rb
@@ -4,6 +4,10 @@ module IndexerClients
   class Prowlarr < Base
     Result = IndexerClients::Result
 
+    # Prowlarr's own default when a definition carries no priority, so
+    # de-duplication always has something to order by.
+    DEFAULT_INDEXER_PRIORITY = 25
+
     class << self
       def search(query, categories: nil, book_type: nil, limit: 100, title: nil, author: nil)
         ensure_configured!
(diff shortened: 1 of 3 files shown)


app/services/indexer_clients/prowlarr.rb
```diff
@@ -202,6 +203,71 @@ def parse_date(date_string)
...
+      # contributes nothing, rather than being queried on the free text.
+      # Most torrent trackers advertise `q` only, so a structured query
+      # retrieves literally zero results there while looking healthy.
+      #
+      # Group the indexers we are about to search by what they can actually
+      # accept and give each group the richest query it supports. A token is
+      # only ever emitted when the indexer is known to support it; everything
+      # else falls back to free text, which every indexer accepts.
+      def book_search_plans(query, title:, author:)
+        structured, free_text = Array(scoped_indexers).partition do |indexer|
```
Pedro-Revez-Silva: **Blocking:** `/api/v1/indexer` returns all configured definitions, while Prowlarr intersects explicit `indexerIds` with `Enabled()` (which also excludes temporarily blocked indexers) and returns HTTP 400 when none of the selected IDs are available. Because every partition becomes an explicit plan and the plans are evaluated fail-fast, a disabled/blocked-only partition aborts the whole book search and can discard valid results from the other partition.

I reproduced this with one enabled structured indexer returning a result and one disabled `q`-only indexer: `ProwlarrClient.search` raised `Prowlarr API error: 400`. This matches [Prowlarr v2.4.0.5397’s dispatch path](https://github.com/Prowlarr/Prowlarr/blob/d6e8466d3ee32915d35476b9c225453984992697/src/NzbDrone.Core/IndexerSearch/ReleaseSearchService.cs#L159-L175).

Please exclude disabled definitions up front and make plan execution tolerate an unavailable partition when another plan succeeds, propagating an error only when every plan fails. A regression test with enabled and disabled/blocked groups would cover this.

app/services/indexer_clients/prowlarr.rb
```diff
@@ -8,22 +8,14 @@ class << self
...
-        params[:indexerIds] = indexer_ids if indexer_ids.present?
 
-        response = request { connection.get("api/v1/search", params) }
+        if search_type_for(title: title, author: author) == "search"
+          return execute_search(query, type: "search", categories: cats, limit: limit, indexer_ids: filtered_indexer_ids)
+        end
 
-        handle_response(response) do |data|
-          Array(data).map { |item| parse_result(item) }
+        book_search_plans(query, title: title, author: author).flat_map do |plan|
```
Pedro-Revez-Silva: **Non-blocking:** concatenating the plan responses bypasses Prowlarr’s global GUID deduplication. If both capability groups return the same GUID, this client returns it twice; on primary-only SearchJob paths, `save_results` then updates the same row twice and the later free-text copy overwrites the earlier copy. Prowlarr normally deduplicates by GUID and selects the higher-priority indexer ([source](https://github.com/Prowlarr/Prowlarr/blob/d6e8466d3ee32915d35476b9c225453984992697/src/NzbDrone.Core/IndexerSearch/ReleaseSearchService.cs#L274-L279)).

Default broad searches and ebook fallbacks already hide most occurrences, so I don’t consider this blocking. Please consider merging the collected plan results by GUID, ideally retaining Prowlarr’s indexer-priority semantics.

Pedro-Revez-Silva (changes requested):
Thanks for the detailed diagnosis—the capability-aware split is the right direction. I’m requesting changes for one correctness issue: an unavailable capability group can currently fail the entire search and discard results returned by another group. I also left a non-blocking note about preserving Prowlarr’s GUID deduplication across the split. I verified the blocking path against Prowlarr v2.4.0.5397 and reproduced it with a focused regression test.

app/services/indexer_clients/prowlarr.rb
coffbr01: You're right, and thank you for catching it — the split turned every partition into an explicit selection, which is a request shape this code never sent before. Fixed in ad4595d.

I reproduced the blocking path on Prowlarr 2.4.0.5397. My own install happens to carry a disabled indexer (1337x, `enable: false`, id 9), so no setup was needed:

```
selection naming only the disabled indexer -> IndexerClients::Base::Error: Prowlarr API error: 400
```

Three changes:

**Disabled definitions no longer shape a plan.** `searchable?` drops `enable: false`, and — where Prowlarr reports it — an indexer whose `status.disabledTill` is still in the future. That second check is belt-and-braces: `IndexerResource` defines `Status` on `develop`, but 2.4.0.5397 does not return the field at all (the live payload keys are `added, appProfileId, capabilities, …, enable, …, priority, …` with no `status`), so blocked indexers are in practice handled by the tolerance below rather than by pre-filtering. If every indexer in scope is unavailable, there is nothing to search and no request is made.

**A single plan sends no selection at all.** If the split leaves one plan, it covers every indexer Prowlarr would have searched anyway, so the `indexerIds` are dropped and the request goes out in exactly the shape it had before this PR. This is the common case — an all-free-text indexer set, which is the configuration the PR exists for — and it means the blocking path is unreachable there, including for an indexer blocked in the window between reading the list and searching. My install now issues one unrestricted free-text query where the previous revision pinned `[9, 11, 12, 10, 7]`, disabled indexer included.

**Plan execution tolerates a failure.** A failing plan is logged and skipped; the error is only propagated when *every* plan fails.

```
a plan naming the disabled indexer + a healthy plan
  -> [IndexerClients::Prowlarr] Book search plan failed (9): Prowlarr API error: 400
  -> 31 results kept
```

Five tests cover this: a disabled indexer never shaping a plan, a blocked one never shaping a plan, no request at all when everything in scope is unavailable, one plan's results surviving another's failure, and the error still propagating when every plan fails. I ran each against the previous revision first — the first four fail there, as does the updated single-plan expectation; the all-plans-fail case is the one that already passed, since the old code raised on the first failure regardless.

app/services/indexer_clients/prowlarr.rb
coffbr01: Agreed, and fixed in ad4595d — the double-write on primary-only `SearchJob` paths is a real consequence, not just a cosmetic duplicate.

`execute_search` now returns raw items and `search` merges them by GUID before parsing, keeping the copy from the indexer with the lowest `priority` and falling back to Prowlarr's own default of 25 when a definition does not carry one. That mirrors `DeDupeReleases`, which groups by GUID and orders by ascending `IndexerPriority`. First-seen order is preserved, and an item without a GUID is keyed on identity so it is never merged into another.

Verified live against Prowlarr 2.4.0.5397 with two deliberately overlapping plans (MyAnonamouse, `priority: 5`, in both):

```
raw items across both plans = 51
distinct guids              = 31
after de-dup                = 31
duplicate guids remaining   = 0
```

31 is also exactly what the unsplit single-plan search returns, so the merge lands on the same result set Prowlarr would have produced itself. A unit test covers the priority tie-break: the same GUID from a `priority: 50` and a `priority: 1` indexer keeps the latter's copy.

app/services/indexer_clients/prowlarr.rb
```diff
@@ -202,6 +210,163 @@ def parse_date(date_string)
...
+
+        plans = []
+        plans << search_plan(build_query(query, title: title, author: author), indexer_ids(structured)) if structured.any?
+        plans << search_plan(free_text_query, indexer_ids(free_text)) if free_text.any?
+
+        # One plan covers every indexer Prowlarr would have searched anyway, so
+        # drop the selection and leave the choice to Prowlarr exactly as before
+        # this split existed. An indexer blocked since the list was read then
+        # costs nothing.
+        plans.first[:indexer_ids] = nil if plans.one? && indexer_filter_tags.empty?
```
Pedro-Revez-Silva: [P2] Preserve the resolved tag scope when collapsing one plan

For name-based `prowlarr_tags`, `scoped_indexers` has already called `indexer_filter_tags` and filtered the indexer list, but this line resolves the names a second time. If the first `GET /api/v1/tag` succeeds and the second transiently returns 500, `configured_tag_ids_for_names` rescues to `[]`, this condition clears the already-correct `indexerIds`, and the subsequent book search is sent unrestricted to every enabled indexer. A focused WebMock repro at this head made the two tag calls (200 then 500) and observed `GET /api/v1/search?...type=book` with no `indexerIds`; the assertion requiring the resolved tagged ID `2` failed (1 test, 2 assertions, 1 failure). Carry the initially resolved filter state/tags into `book_search_plans` and use that stable value here instead of re-resolving names.

Pedro-Revez-Silva (changes requested):
Reviewed head ad4595d989d4649733554e12ec2d63e4733874fe against merge-base f60580ff6d237fe4479c1b21f2cb8e5e6459187e. The follow-up fixes the previously reported unavailable-plan and GUID-dedup paths, and the targeted Prowlarr/SearchJob suites pass (118 tests, 396 assertions). One P2 remains: a transient failure in the new second tag-name resolution can drop a configured Prowlarr tag scope and issue an unrestricted book search. Requesting changes until the already-resolved scope is preserved.

app/services/indexer_clients/prowlarr.rb
coffbr01: Confirmed and fixed in 2835bf7 — thank you, this is a worse failure than a dropped search: the scope silently widens to every enabled indexer, which is the one thing `prowlarr_tags` exists to prevent.

I reproduced your repro before changing anything. With `prowlarr_tags = "books"` (name-based), `/api/v1/tag` answering 200 once and 500 thereafter, and two `q`-only indexers (id 1 untagged, id 2 tagged `books`):

```
GET /api/v1/search?categories=7030&limit=100&query=Frankenstein%20Mary%20Shelley&type=book
  -> no indexerIds
```

The collapse path made it reachable exactly as you describe: `scoped_indexers` filtered the list correctly on the first resolution, the split left one plan, and the second `indexer_filter_tags` call rescued to `[]`, so `plans.one? && indexer_filter_tags.empty?` cleared a correct selection.

The fix carries the state rather than re-deriving it. `search` resolves the scope once and threads it through:

```ruby
tags = indexer_filter_tags
scoped = scoped_indexers(tags: tags)
plans = book_search_plans(query, title: title, author: author, tags: tags, indexers: scoped)
```

`book_search_plans` takes `tags:` (defaulting to `indexer_filter_tags`, with `indexers:` defaulting to `scoped_indexers(tags: tags)`, so a standalone call still resolves both from one value) and the collapse tests that parameter. There is now exactly one tag resolution per search, so the two calls can no longer disagree — and the comment there now says why the collapse is only sound without a scope: with one, "every indexer Prowlarr would have searched" is the enabled set, not the tagged subset.

The regression test asserts the search still carries `indexerIds=2` when the tag endpoint fails on the second call. I ran it against ad4595d first and it fails there with the unrestricted request above, so it pins the behaviour rather than the implementation.

Targeted suites pass on Ruby 3.3.6 (`test/services/prowlarr_client_test.rb` + `test/jobs/search_job_test.rb`): 119 runs, 397 assertions, 0 failures — your 118/396 plus this test. Rubocop clean on both changed files.

One thing I deliberately left alone: if the *first* resolution fails, the scope is still lost and the search goes out unrestricted. That path predates this PR — `filtered_indexer_ids` has always rescued a tag-lookup failure to "no filter" on the `type=search` path — so treating a tag outage as fail-closed (or surfacing it) looks like its own change to me, and I did not want to smuggle a behaviour change into a regression fix. Happy to open a separate PR for it if you'd like it fixed.

Pedro-Revez-Silva (approved):
Re-reviewed exact head `2835bf7bbdcf4d9073501afd5012ae9cac22ae35`. The prior tag-scope race is fixed by resolving the configured scope once and threading it through planning. The regression passes on this head and reproduces the old failure on `ad4595d`; targeted tests (119 runs, 397 assertions), RuboCop, quality, security, validation, and Docker checks are all green.

PR #410, opened 2026-07-29
feat: attended watched-folder import

AlexanderReaper7:
This adds a watched-folder import feature to allow preexisting files to be imported without uploading through shelfarr.
I needed this because i have a large amount of torrents that i wanted to import into a managed library, and since shelfarr is taking the role that both radarr and seer takes for movies but for books, that falls unto shelfarr. 

Video stack | Books stack | Role
-- | -- | --
Seerr / Jellyseerr | Shelfarr | Request + acquire (search indexers, download, deliver)
Radarr / Sonarr | (baked into Shelfarr) | Grab + import automation
Jellyfin | Audiobookshelf | Library server — organize, browse, stream/read

These are the relevant pages:

| Target Page | Relative Route / Fragment | Description |
| --- | --- | --- |
| **Imports** | `/admin/detected_imports` | Page route |
| **Settings** | `/admin/settings#Downloads:~:text=Watched%20Folder%20Import` |  *Watched Folder Import* section |

>[!Note]
>I don't speak ruby so,
>**The feature and the rest of the PR is written by Opus**

---

## Suggested reading order for the reviewer

1. **[config/routes.rb:159-171](/config/routes.rb#L159-L171)** — the routes above.
2. **[Admin::DetectedImportsController](/app/controllers/admin/detected_imports_controller.rb)** — thin actions; `index` (lazy imported-history), `import`→`apply_selection!`, `scan`, `undo`.
3. **[LibraryAcquisitionService](/app/services/library_acquisition_service.rb)** — `identify` (read-only ranking), `import!` (reserve → import → record publication → claim `file_path` → scan), and `undo_import!` / `reverse_publication!`. The heart of it.
4. **[LibraryFileImporter](/app/services/library_file_importer.rb)** — copy/move/hardlink dispatch over `FileCopyService`, plus the publication record it returns (the safety-critical bit).
5. **[WatchedFolderScanService](/app/services/watched_folder_scan_service.rb)** — the walk + candidate classification (plain / multi-disc / collection).
6. **[DetectedImport](/app/models/detected_import.rb)** + **[migration](/db/migrate/20260722120000_create_detected_imports.rb)** — statuses and the `(device, inode)` idempotency index.

---

## Summary
- add an **attended** watched-folder importer (behind a default-off `library_import_enabled` setting) that detects pre-existing book files which were not acquired through a Shelfarr request and queues them for admin review before importing — filling the gap where Shelfarr replaces both requester and acquirer but, unlike Radarr/Sonarr, can't adopt files it didn't download
- group candidates the way the download importer sees releases: a single ebook/comic file, a self-contained audiobook folder, a multi-disc book (CD/Disc subfolders collapse to one import), or a collection/box set (each titled subfolder becomes its own import)
- import approved items into the organised library using the existing `completed_download_import_mode` (copy/move/hardlink), reusing `FileCopyService`'s atomic, no-follow, TOCTOU-safe primitives; the source is never mutated except by an explicit move, and a failed import leaves it untouched
- anchor idempotency on the source `(device, inode)` pair rather than the path, so hardlink imports and re-scans never re-detect the same content
- record exactly what each import published and offer a **reversible undo**, so an item approved against the wrong match can be returned to the review queue and re-imported (see below)
- add the admin review queue (live Turbo morph-refresh, manual "Scan now" with in-progress/last-result feedback, and an expand/collapse imported-history section that lazy-loads the full history) plus a batched `import_detected` admin notification wired into the webhook/Discord event allow-list

## Reversible import (publication record + undo)

An attended importer needs an undo — the whole point of the review step is that a human can be wrong about a match, and without a way back the only remedy is hand-editing the library. Making that safe is the bulk of the code added since the first review pass.

- every import writes a **publication record** onto the detection (one additive migration, `publication_record` json) listing every file it created and every directory it brought into existence, with each entry's `(device, inode)`. It is persisted *before* the database claim, so a crash or a lost claim race is still precisely reversible: the bytes are already on disk at that point and only an exact record of them makes the rollback safe.
- undo reverses exactly those entries. A copy/hardlink import discards the redundant library artifact; a move import returns each file to the path the scanner recorded, rebuilding a moved tree as it was found. Directories are removed only if this import created them **and** they are still empty — a templated `Author/Book` directory is routinely shared with files this import does not own, so it is never deleted recursively.
- containment is structural rather than a pathname comparison: every removal walks from the output root through pinned `O_NOFOLLOW` descriptors and is gated on the recorded `(device, inode)`, so neither an ancestor swapped for a symlink nor a root that canonicalizes differently (`/var` vs `/private/var`) can redirect or block it.
- undo resets the detection and un-acquires the book **only when the reversal actually completed**. A library file that is no longer the one this import published (renamed, re-tagged, replaced) is left alone and the undo raises, rather than reporting success while stranding an artifact a re-import would then duplicate. The record is kept until that state reset commits, so a failure (or a process death) in that window still leaves something to retry from.
- returning a moved file is gated on identity at both ends. Each entry records the source `(device, inode)` taken *before* publication, so undo can tell "the original is still there" from "unrelated bytes have taken the path over" — the second is a conflict, not a licence to delete the library's only copy. The restore itself walks the recorded watched-folder root through pinned no-follow descriptors, so a source ancestor swapped for a symlink cannot redirect it.
- an import that finds a journal from an interrupted attempt reconciles it before running: a publication that already claimed the book is handed back rather than repeated, and a stranded one is reversed so the retry imports from the restored source instead of a path a move consumed.
- an import that predates publication tracking refuses to undo rather than guess which files under the templated directory were its own.
- the `(device, inode)` recorded at detection is now carried into the importer, which refuses to publish a source whose inode changed between approval and import — a path swapped in that window cannot substitute different bytes under an approved title.

Two `FileCopyService` primitives grew optional parameters to support this, both backward-compatible: `mv_noreplace` accepts a caller-supplied `source_snapshot` (so a tree import can pin one immutable snapshot for the whole release), and `remove_directory_child_if_identity` accepts `expected_entries` (so a caller that only owns an *empty* directory can pass `{}` and have the removal restore the directory untouched if anything has since been written into it). Omitting either keeps the previous behaviour.

## Design notes
- a human gates every import, so the matcher only produces a ranked suggestion (embedded metadata → filename → online) and never has to be autonomously correct — this removes the confidence-threshold policy and the risk of silently mis-tagging the library
- the import engine is a new `LibraryFileImporter` that delegates every filesystem publication to `FileCopyService`, rather than extracting the core out of `UploadProcessingJob`/`PostProcessingJob`. This leaves the most safety-critical request pipeline untouched (zero regression risk) and confines all new code to a default-off feature. Unifying both importers behind one engine is a reasonable follow-up once this is proven in CI.
- a multi-file release is refused when its book type has no path template, instead of scattering the tracks loose into the output root and claiming the root itself as `book.file_path`. `UploadZipImportFileService` already refuses the same configuration.
- ships behind `library_import_enabled` (default off) with two additive migrations (`create_detected_imports`, `add_publication_record_to_detected_imports`); existing installs are unaffected
- the one change outside the feature's own files is in `PostProcessingJob`: the scanner queues anything under the library import path, which commonly includes the download client's completed folder, so a download the scanner reached first would sit in the review queue while post-processing imports it — and approving that row afterwards would import the title twice. Those rows are now retired once the content is in the library.

## Test plan
- focused feature suites (Docker, `RAILS_ENV=test`): `DetectedImport` model, `LibraryAcquisitionService` (incl. journal-write abort, undo-transaction rollback and retry, reoccupied source path, swapped source ancestor, interrupted-publication reversal, resumed claim), scanner (incl. collection-split and multi-disc grouping, known-item skipping, symlink refusal, output-root overlap rejection, stale-inode release), scan job, import job, and controller — **69 runs, 244 assertions, 0 failures**
- full `bin/rails test` in Docker — **2961 runs, 11045 assertions, 0 failures, 0 errors**. The 10 pre-existing failures reported on the previous revision of this PR were an artifact of running the suite over a Docker-on-Windows bind mount, which lacks atomic publication (`linkat`/`renameat2(RENAME_NOREPLACE)`) and extended attributes (the environmental limitation noted in #379); copying the source into the container's overlayfs instead, the suite is clean.
- RuboCop (`--fail-level warning`) on all new/changed Ruby files — no offenses
- not covered by a unit test: the `scan_job_pending?` failed-execution exclusion. Solid Queue's tables live in a separate database that the test suite does not prepare, which is also why the identical scopes on `DownloadMonitorJob` and friends are untested.
- not run in this environment (no Ruby toolchain on the Windows host — everything above ran in the app Docker image): Brakeman, system/browser tests, coverage, and mutant. Please run the full `bin/quality push` / `bin/quality deep` gate in CI before merging.

### DIFF
--- a/app/jobs/detected_import_job.rb
+++ b/app/jobs/detected_import_job.rb
@@ -1,0 +1,75 @@
+# frozen_string_literal: true
+
+# Imports an admin-approved DetectedImport into the organised library via
+# LibraryAcquisitionService. Claims the record with a compare-and-swap so a
+# double submit or redelivery cannot run two imports for the same detection.
+# A failed import leaves the source untouched and is retryable from the review
+# queue.
+class DetectedImportJob < ApplicationJob
+  queue_as :default
+
+  def perform(detected_import_id)
+    detected_import = claim(detected_import_id)
+    return unless detected_import
+
+    begin
+      book = detected_import.suggested_book || create_and_attach_book(detected_import)
+
+      result = LibraryAcquisitionService.import!(
+        source_path: detected_import.source_path,
+        book: book,
+        owner: detected_import,
+        provenance: :watched_folder
+      )
+
+      detected_import.update!(
+        status: "imported",
+        imported_book: result.book,
+        suggested_book: result.book,
+        error_message: nil
+      )
+      Rails.logger.info(
+        "[DetectedImportJob] Imported detection ##{detected_import.id} into book ##{result.book.id}"
+      )
+    rescue => e
+      Rails.logger.error(
+        "[DetectedImportJob] Import failed for detection ##{detected_import.id} (#{e.class})"
+      )
+      detected_import.update!(
+        status: "failed",
+        error_message: e.message.to_s.scrub.truncate(2_000)
+      )
+    end
+  end
+
+  private
+
+  # Compare-and-swap detected/failed (or an abandoned, stale "importing" whose
+  # worker died before finishing) -> importing, so only one worker proceeds and
+  # a wedged row can be recovered.
+  def claim(detected_import_id)
+    claimed = DetectedImport
+      .where(id: detected_import_id)
+      .where(
+        "status IN (:actionable) OR (status = 'importing' AND updated_at < :stuck_before)",
+        actionable: DetectedImport::ACTIONABLE_STATUSES,
+        stuck_before: DetectedImport::STUCK_IMPORTING_AFTER.ago
+      )
+      .update_all(status: "importing", updated_at: Time.current)
+    return unless claimed == 1
+
+    DetectedImport.find_by(id: detected_import_id)
+  end
+
+  # Persist the created Book onto the detection so a retry after a failed import
+  # reuses it instead of orphaning a new unacquired record each attempt.
+  def create_and_attach_book(detected_import)
+    book = Book.create!(
+      title: detected_import.parsed_title.presence || File.basename(detected_import.source_path.to_s),
+      author: detected_import.parsed_author,
+      book_type: detected_import.book_type.presence || "ebook"
+    )
+    detected_import.update!(suggested_book: book)
+    book
+  end
+end
--- a/app/jobs/detected_import_job.rb
+++ b/app/jobs/detected_import_job.rb
@@ -20,7 +20,8 @@ class DetectedImportJob < ApplicationJob
         book: book,
         owner: detected_import,
         provenance: :watched_folder,
-        source_identity: detected_import.source_identity
+        source_identity: detected_import.source_identity,
+        source_base: WatchedFolderScanService.import_root
       )
 
       detected_import.update!(
--- a/app/jobs/detected_import_job.rb
+++ b/app/jobs/detected_import_job.rb
@@ -19,7 +19,8 @@ class DetectedImportJob < ApplicationJob
         source_path: detected_import.source_path,
         book: book,
         owner: detected_import,
-        provenance: :watched_folder
+        provenance: :watched_folder,
+        source_identity: detected_import.source_identity
       )
 
       detected_import.update!(
--- a/app/jobs/detected_import_job.rb
+++ b/app/jobs/detected_import_job.rb
@@ -8,6 +8,20 @@
 class DetectedImportJob < ApplicationJob
   queue_as :default
 
+  # One import at a time per detection. The compare-and-swap below stops two
+  # workers claiming the same row, but a row wedged past STUCK_IMPORTING_AFTER is
+  # deliberately re-claimable — and without this lease a genuinely slow import
+  # would still be running when the admin recovers it, letting the second worker
+  # reverse what the first is finalizing. The lease serialises them instead.
+  #
+  # Its duration matches the stuck window on purpose: past that point the system
+  # treats the claiming worker as dead, so the lease must expire with it or a
+  # killed worker would block its own recovery for an hour.
+  limits_concurrency(
+    key: ->(detected_import_id) { "detected_import/#{detected_import_id}" },
+    duration: DetectedImport::STUCK_IMPORTING_AFTER
+  )
+
   def perform(detected_import_id)
     detected_import = claim(detected_import_id)
     return unless detected_import
@@ -46,17 +60,11 @@ class DetectedImportJob < ApplicationJob
 
   private
 
-  # Compare-and-swap detected/failed (or an abandoned, stale "importing" whose
-  # worker died before finishing) -> importing, so only one worker proceeds and
-  # a wedged row can be recovered.
+  # Compare-and-swap claimable -> importing, so only one worker proceeds while a
+  # wedged row stays recoverable.
   def claim(detected_import_id)
-    claimed = DetectedImport
+    claimed = DetectedImport.claimable
       .where(id: detected_import_id)
-      .where(
-        "status IN (:actionable) OR (status = 'importing' AND updated_at < :stuck_before)",
-        actionable: DetectedImport::ACTIONABLE_STATUSES,
-        stuck_before: DetectedImport::STUCK_IMPORTING_AFTER.ago
-      )
       .update_all(status: "importing", updated_at: Time.current)
     return unless claimed == 1
 
--- a/app/jobs/watched_folder_scan_job.rb
+++ b/app/jobs/watched_folder_scan_job.rb
@@ -1,0 +1,148 @@
+# frozen_string_literal: true
+
+# Recurring, self-rescheduling job that scans the configured watched-folder
+# import path for pre-existing book files. Mirrors the scheduling pattern of
+# DownloadMonitorJob / TelegramPollingJob: a single concurrency key prevents
+# overlapping scans, and each run re-arms the next at the configured interval.
+class WatchedFolderScanJob < ApplicationJob
+  SCHEDULE_CACHE_KEY = "watched_folder_scan/next_run_at"
+  # Progress/result of the most recent manual scan, for the review-queue UI.
+  # Backed by the shared cache (solid_cache in production) and written only by
+  # the worker running the scan, so the web process rendering the queue sees a
+  # consistent state and never a stuck "running" flag from a crossed process.
+  STATUS_CACHE_KEY = "watched_folder_scan/status"
+  DEFAULT_INTERVAL_SECONDS = 300
+  MIN_INTERVAL_SECONDS = 30
+  MAX_INTERVAL_SECONDS = 86_400
+
+  queue_as :default
+  limits_concurrency key: "watched_folder_scan", duration: 1.hour, on_conflict: :discard
+
+  class << self
+    def scanning_enabled?
+      SettingsService.get(:library_import_enabled, default: false) &&
+        SettingsService.get(:library_import_path).to_s.strip.present?
+    end
+
+    # Snapshot of the latest manual scan for the queue UI: state ("running" /
+    # "idle"), when it completed, and how many candidates/new detections it saw.
+    # Empty until the first manual scan runs (or in an environment whose cache is
+    # not shared across processes, e.g. development's memory store).
+    def scan_status
+      Rails.cache.read(STATUS_CACHE_KEY) || {}
+    end
+
+    def scanning_now?
+      scan_status[:state] == "running"
+    end
+
+    def mark_running!
+      write_status(state: "running", started_at: Time.current)
+    end
+
+    def mark_completed!(result)
+      write_status(
+        state: "idle",
+        completed_at: Time.current,
+        scanned: result&.scanned,
+        detected: result&.detected,
+        failed: result.nil?
+      )
+    end
+
+    def broadcast_queue_refresh
+      Turbo::StreamsChannel.broadcast_refresh_to(DetectedImport::INDEX_STREAM)
+    rescue => e
+      Rails.logger.warn "[WatchedFolderScanJob] Could not broadcast queue refresh (#{e.class})"
+    end
+
+    def ensure_running!
+      return unless scanning_enabled?
+      return if scan_job_pending?
+
+      next_run_at = Rails.cache.read(SCHEDULE_CACHE_KEY).to_i
+      return if next_run_at > Time.current.to_i
+
+      reserve_schedule!
+      Rails.logger.info "[WatchedFolderScanJob] Scheduling watched-folder scan chain"
+      perform_later
+    end
+
+    def clear_schedule!
+      Rails.cache.delete(SCHEDULE_CACHE_KEY)
+    end
+
+    def scan_job_pending?(excluding_active_job_id: nil)
+      return false unless solid_queue_adapter?
+
+      scope = SolidQueue::Job.where(class_name: name, finished_at: nil)
+      scope = scope.where.not(active_job_id: excluding_active_job_id) if excluding_active_job_id.present?
+      scope.exists?
+    rescue ActiveRecord::StatementInvalid, NameError
+      false
+    end
+
+    def interval_seconds
+      SettingsService.get(:library_import_scan_interval, default: DEFAULT_INTERVAL_SECONDS)
+        .to_i
+        .clamp(MIN_INTERVAL_SECONDS, MAX_INTERVAL_SECONDS)
+    end
+
+    private
+
+    def write_status(attrs)
+      Rails.cache.write(STATUS_CACHE_KEY, attrs, expires_in: 1.day)
+    rescue => e
+      Rails.logger.warn "[WatchedFolderScanJob] Could not record scan status (#{e.class})"
+    end
+
+    def reserve_schedule!
+      Rails.cache.write(
+        SCHEDULE_CACHE_KEY,
+        interval_seconds.seconds.from_now.to_i,
+        expires_in: [ interval_seconds * 3, 300 ].max.seconds
+      )
+    end
+
+    def solid_queue_adapter?
+      ActiveJob::Base.queue_adapter.class.name == "ActiveJob::QueueAdapters::SolidQueueAdapter"
+    end
+  end
+
+  # manual: true is passed by the "Scan now" button. Such a scan announces its
+  # progress to the review queue (a spinner while it runs, its result when it
+  # finishes) and refreshes the queue on completion so new detections appear
+  # without a manual reload. The recurring background scan stays silent — it only
+  # records its completion so "last scanned" stays fresh, and relies on the
+  # per-record broadcasts to surface anything it finds.
+  def perform(manual: false)
+    unless self.class.scanning_enabled?
+      self.class.clear_schedule!
+      return
+    end
+
+    if manual
+      self.class.mark_running!
+      self.class.broadcast_queue_refresh
+    end
+
+    result = WatchedFolderScanService.scan!
+    self.class.mark_completed!(result)
+    self.class.broadcast_queue_refresh if manual
+  rescue => e
+    Rails.logger.error "[WatchedFolderScanJob] Scan failed (#{e.class}): #{e.message}"
+    self.class.mark_completed!(nil)
+    self.class.broadcast_queue_refresh if manual
+  ensure
+    schedule_next_run if self.class.scanning_enabled?
+  end
+
+  private
+
+  def schedule_next_run
+    return if self.class.scan_job_pending?(excluding_active_job_id: job_id)
+
+    self.class.send(:reserve_schedule!)
+    WatchedFolderScanJob.set(wait: self.class.interval_seconds.seconds).perform_later
+  end
+end
--- a/app/jobs/watched_folder_scan_job.rb
+++ b/app/jobs/watched_folder_scan_job.rb
@@ -14,6 +14,11 @@ class WatchedFolderScanJob < ApplicationJob
   DEFAULT_INTERVAL_SECONDS = 300
   MIN_INTERVAL_SECONDS = 30
   MAX_INTERVAL_SECONDS = 86_400
+  # How long the status cache is allowed to claim a scan is running. Matches the
+  # concurrency lease below: past it the claiming worker cannot still hold the
+  # key, so a status stuck on "running" is the residue of a worker that was
+  # killed mid-scan and must not keep disabling "Scan now" for the cache TTL.
+  RUNNING_STATUS_TTL = 1.hour
 
   queue_as :default
   limits_concurrency key: "watched_folder_scan", duration: 1.hour, on_conflict: :discard
@@ -29,7 +34,14 @@ class WatchedFolderScanJob < ApplicationJob
     # Empty until the first manual scan runs (or in an environment whose cache is
     # not shared across processes, e.g. development's memory store).
     def scan_status
-      Rails.cache.read(STATUS_CACHE_KEY) || {}
+      status = Rails.cache.read(STATUS_CACHE_KEY) || {}
+      return status unless stale_running?(status)
+
+      # Report the abandoned run as a failed scan rather than a live one, so the
+      # queue offers "Scan now" again instead of spinning until the cache
+      # expires. Left in the cache as-is: the next scan overwrites it, and a
+      # write here would race the worker that owns the key.
+      status.merge(state: "idle", completed_at: status[:started_at], failed: true)
     end
 
     def scanning_now?
@@ -89,6 +101,11 @@ class WatchedFolderScanJob < ApplicationJob
     end
 
     private
+
+    def stale_running?(status)
+      status[:state] == "running" &&
+        (status[:started_at].blank? || status[:started_at] < RUNNING_STATUS_TTL.ago)
+    end
 
     def write_status(attrs)
       Rails.cache.write(STATUS_CACHE_KEY, attrs, expires_in: 1.day)
--- a/app/jobs/watched_folder_scan_job.rb
+++ b/app/jobs/watched_folder_scan_job.rb
@@ -6,18 +6,16 @@
 # overlapping scans, and each run re-arms the next at the configured interval.
 class WatchedFolderScanJob < ApplicationJob
   SCHEDULE_CACHE_KEY = "watched_folder_scan/next_run_at"
-  # Progress/result of the most recent manual scan, for the review-queue UI.
-  # Backed by the shared cache (solid_cache in production) and written only by
-  # the worker running the scan, so the web process rendering the queue sees a
-  # consistent state and never a stuck "running" flag from a crossed process.
+  # Progress/result of the latest scan, for the review-queue UI. Backed by the
+  # shared cache (solid_cache in production) and written only by the worker
+  # running the scan, so the web process never sees a crossed-process state.
   STATUS_CACHE_KEY = "watched_folder_scan/status"
   DEFAULT_INTERVAL_SECONDS = 300
   MIN_INTERVAL_SECONDS = 30
   MAX_INTERVAL_SECONDS = 86_400
-  # How long the status cache is allowed to claim a scan is running. Matches the
-  # concurrency lease below: past it the claiming worker cannot still hold the
-  # key, so a status stuck on "running" is the residue of a worker that was
-  # killed mid-scan and must not keep disabling "Scan now" for the cache TTL.
+  # How long the cache may claim a scan is running. Matches the concurrency lease
+  # below: past it the worker cannot still hold the key, so a stuck "running" is
+  # the residue of a killed worker and must not keep disabling "Scan now".
   RUNNING_STATUS_TTL = 1.hour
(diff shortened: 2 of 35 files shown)


app/services/library_file_importer.rb
```diff
@@ -0,0 +1,170 @@
...
+    rescue Errno::EEXIST
+      counter += 1
+      retry
+    end
+  end
+
+  def publish_with_mode(source, destination, source_root:)
+    case @mode
+    when "move"
+      FileCopyService.mv_noreplace(
```
Pedro-Revez-Silva: [P1] Directory imports in `move` mode fail after consuming the first source file. `import_directory` snapshots the tree and passes that snapshot into each `mv_noreplace`; after the first move unlinks its source, `with_pinned_source_root` revalidates the old manifest and raises `Errno::ESTALE`. I reproduced this with an ordinary two-MP3 directory: chapter 1 disappeared from the watched folder and appeared in the library, then the import failed before chapter 2. Use a directory-level move or defer/coordinate source removal so a successful per-file publication does not invalidate the snapshot mid-import.

app/services/library_acquisition_service.rb
```diff
@@ -0,0 +1,422 @@
...
+          book: book,
+          base_path: base_path
+        )
+        claim_file_path!(book, result.imported_path, owner)
+        trigger_library_scan(book)
+        Rails.logger.info(
+          "[LibraryAcquisitionService] Imported #{provenance || 'source'} for book ##{book.id} (mode=#{mode})"
+        )
+        ImportResult.new(book: book, destination_path: result.imported_path, mode: mode)
+      rescue
```
Pedro-Revez-Silva, line 165: [P1] Roll back the publication when finalizing the database claim fails. The importer has already copied/moved bytes before `claim_file_path!`; this rescue only clears the reservation. I forced the supported CAS-conflict path in `move` mode and the source was gone while an unclaimed EPUB remained in the library and the book had no `file_path`. A retry can then create a numbered duplicate. Preserve an exact publication record and reverse/reconcile it before releasing the reservation.

app/services/library_acquisition_service.rb
```diff
@@ -0,0 +1,422 @@
...
+
+      unless within_output_root?(destination, book)
+        raise AcquisitionConflictError,
+          "Refusing to undo: #{destination} is not inside the library output path"
+      end
+
+      if File.exist?(detected_import.source_path)
+        # Copy or hardlink: the watched-folder source is still present, so the
+        # library artifact is a redundant second copy — discard it.
+        FileUtils.remove_entry(destination)
```
Pedro-Revez-Silva: [P1] Do not recursively delete the whole destination directory unless this import exclusively created and owns it. Directory imports merge files into an already-existing templated directory, but undo records only the directory path and removes everything beneath it. I placed `preexisting.txt` in `Author/Book`, imported an audiobook there, then undid it; the unrelated file was deleted. Track and remove only this import's published entries, or require an exclusive new destination directory.

app/services/library_acquisition_service.rb
```diff
@@ -0,0 +1,422 @@
...
+      else
+        inner = Dir.glob(File.join(destination, "**", "*")).find { |path| File.file?(path) }
+        FileUtils.mv(inner || destination, source)
+        FileUtils.remove_entry(destination) if File.directory?(destination)
+      end
+    end
+
+    def within_output_root?(destination, book)
+      base = File.realpath(output_base_path(book))
+      target = File.expand_path(destination)
```
Pedro-Revez-Silva: [P1] This containment check does not make the subsequent `FileUtils.remove_entry` safe. `base` is canonicalized while `target` is only lexical: after replacing `output/Author` with a symlink to an external directory, the prefix check still passed and undo deleted `outside/Book/victim.txt`. Conversely, a legitimate output reached through `/var` on macOS is rejected because the base becomes `/private/var` while the target stays `/var` (the PR's undo tests fail this way). Resolve and operate through pinned no-follow descriptors, rather than checking one pathname and deleting through another resolution.

app/jobs/detected_import_job.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  def perform(detected_import_id)
+    detected_import = claim(detected_import_id)
+    return unless detected_import
+
+    begin
+      book = detected_import.suggested_book || create_and_attach_book(detected_import)
+
+      result = LibraryAcquisitionService.import!(
+        source_path: detected_import.source_path,
```
Pedro-Revez-Silva, line 33: [P1] Revalidate the source identity captured at detection before importing. The scanner persists `source_device`/`source_inode`, but this job uses only the mutable pathname. I detected one EPUB, atomically replaced that path with a different inode and contents, and ran the job; Shelfarr imported the replacement bytes under the approved title and marked the detection `imported`. Compare the current pinned source to the stored identity (and fingerprint if in-place changes must also be rejected) before publication.

Pedro-Revez-Silva (changes requested):
Requesting changes for five independently reproduced filesystem/data-integrity issues on this exact head. The reproductions cover an ordinary two-file move, a post-publication claim conflict, undo with pre-existing files, an ancestor symlink swap, and replacement of an approved source inode.

app/services/library_acquisition_service.rb
```diff
@@ -0,0 +1,607 @@
...
+    end
+
+    # Persist what an import published so a later rollback or undo can reverse
+    # exactly those entries. Written with update_columns so an in-flight import
+    # neither runs validations nor broadcasts a status change.
+    def record_publication!(owner, publication)
+      return unless owner.respond_to?(:publication_record=) && owner.persisted?
+
+      owner.update_columns(publication_record: publication, updated_at: Time.current)
+    rescue => e
```
Pedro-Revez-Silva: [P1] Abort when the publication journal cannot persist

This journal is the only durable record that can reverse files already published by a destructive move, but this rescue converts its database failure into a successful import. I forced this `update_columns` call to fail once: the source was consumed, the book was acquired with a library `file_path`, and the detection completed with `publication_record = nil`, so undo refused and there was no safe rollback record. Let the write failure propagate before claiming the book, and reverse the in-memory publication record rather than committing an unjournaled import.

app/services/library_acquisition_service.rb
```diff
@@ -0,0 +1,607 @@
...
+      root = publication["base_path"].presence
+      return files.empty? if root.blank?
+
+      move = publication["mode"].to_s == "move"
+      complete = files.reverse_each.map { |entry|
+        move ? restore_moved_file!(entry, root) : discard_published_file!(entry, root)
+      }.all?
+      Array(publication["directories"]).reverse_each { |entry| discard_created_directory!(entry, root) }
+
+      clear_publication_record!(owner) if complete
```
Pedro-Revez-Silva: [P1] Keep the publication journal until undo state commits

A complete filesystem reversal clears `publication_record` here before `undo_import!` begins its database transaction. I forced `release_book_after_undo!` to fail: the library file was already removed, but the transaction rolled back to an `imported` detection and acquired book with a stale `file_path`; the journal was nil, so every retry refused to undo. A process death in this window has the same result. Retain a durable undo/recovery phase until the detection and book reset commits, then clear the journal.

app/services/library_acquisition_service.rb
```diff
@@ -0,0 +1,607 @@
...
+    def restore_moved_file!(entry, root)
+      destination = entry["destination"].presence
+      source = entry["source"].presence
+      return true if destination.blank? || source.blank?
+
+      # A directory move publishes every file before unlinking the source tree,
+      # so a reversal partway through finds the source still in place. The
+      # library copy is then redundant rather than the only surviving one. The
+      # same branch makes a retried undo a no-op for entries already returned.
+      return discard_published_file!(entry, root) if File.exist?(source)
```
Pedro-Revez-Silva: [P1] Refuse undo when a moved source path has been reused

`File.exist?(source)` only proves that something occupies the old pathname, not that the imported bytes survived there. I move-imported an EPUB, created unrelated bytes at the now-free source path, and ran undo; this branch deleted the only original bytes from the library, retained the unrelated source, and reset the detection successfully. Treat an occupied source as a conflict unless its content/identity is proven equivalent; otherwise restore with no-replace semantics and leave the library artifact and state intact.

app/services/library_acquisition_service.rb
```diff
@@ -0,0 +1,607 @@
...
+
+      unless published_file_state(entry, root) == :match
+        Rails.logger.warn(
+          "[LibraryAcquisitionService] Could not return #{destination} to #{source}: " \
+            "it is no longer the file this import published"
+        )
+        return false
+      end
+
+      FileUtils.mkdir_p(File.dirname(source))
```
Pedro-Revez-Silva: [P1] Restore moved files through the pinned watched-folder root

This creates the stored source parent with pathname-following `mkdir_p`, then trusts that mutable parent as the destination root. I replaced a descendant of the watched folder with a symlink to an external directory before undo; Shelfarr copied the book through the symlink, deleted the library copy, and reported success. Resolve the recorded source relative to the configured watched-folder root and create/publish through pinned no-follow descriptors, refusing changed source ancestry.

app/services/library_acquisition_service.rb
```diff
@@ -0,0 +1,607 @@
...
+      importer = LibraryFileImporter.new(mode: mode)
+      claimed = false
+      begin
+        result = importer.import(
+          source: source_path,
+          book: book,
+          base_path: base_path,
+          expected_source_identity: source_identity
+        )
+        record_publication!(owner, result.publication)
```
Pedro-Revez-Silva, line 148: [P2] Reconcile an interrupted durable publication before retrying

This creates a durable intermediate state after the source may already be consumed but before the book claim and final detection status. I simulated worker death immediately after publication persistence, expired the lease, and retried the job: the retry attempted a fresh import from the missing source and ended `failed`, while the library bytes, publication journal, and book reservation remained stranded and undo was unavailable. On a stale claim, detect this same-owner phase and resume claim/finalization or safely reverse it instead of rerunning from the source.

app/views/admin/detected_imports/index.html.erb
```diff
@@ -0,0 +1,194 @@
...
+                  <%= item.status.titleize %>
+                </span>
+                <% if item.status == "failed" && item.error_message.present? %>
+                  <div class="text-red-400 text-xs mt-1 break-words max-w-xs"><%= item.error_message %></div>
+                <% end %>
+              </td>
+              <td class="px-6 py-4 whitespace-nowrap text-right text-sm">
+                <% if item.actionable? %>
+                  <div class="flex items-center justify-end gap-2">
+                    <%= button_to "Import", import_admin_detected_import_path(item), method: :post,
```
Pedro-Revez-Silva, line 105: [P2] Preserve the displayed match in one-click imports

This row can display `best_candidate` as its Suggested match, but the one-click form submits no `selection`. `apply_selection!` therefore leaves `suggested_book` nil and the job creates a new Book from the filename parse instead of resolving what the admin approved. I reproduced this with a displayed `hardcover:410123` candidate scoring 99: Import created the deliberately wrong parsed title with no `unified_work_id`. Submit `item.default_selection`, or require Review whenever no persisted `suggested_book` exists.

app/jobs/watched_folder_scan_job.rb
```diff
@@ -0,0 +1,165 @@
...
+    end
+
+    def clear_schedule!
+      Rails.cache.delete(SCHEDULE_CACHE_KEY)
+    end
+
+    def scan_job_pending?(excluding_active_job_id: nil)
+      return false unless solid_queue_adapter?
+
+      scope = SolidQueue::Job.where(class_name: name, finished_at: nil)
```
Pedro-Revez-Silva: [P2] Ignore failed jobs when restarting the scan chain

Solid Queue retains failed jobs with `finished_at = NULL`, so this scope treats a failed execution as pending forever. I reproduced a failed `WatchedFolderScanJob` row making `scan_job_pending?` return true; subsequent boot/settings calls to `ensure_running!` therefore never enqueue a replacement. Exclude rows with a failed execution (as the download monitor does) or explicitly reconcile/retry failed chain jobs.

Pedro-Revez-Silva (changes requested):
Re-reviewed exact head eddcafc54c75e4e0a9e382de330f21ee75185930 against merge-base fdd8dd05f8c984b8980f70e99525999b8497d151 (current base tip 2b6b6f7856ca2427df61108411af3a7b80ab0710). The original two-file move, claim-conflict rollback, shared-destination undo, ancestor-swap, and different-inode source-replacement reproductions now pass. However, the existing journal/atomic-removal/content-identity threads remain actionable in narrower failure paths, and seven independent recovery/UI/scheduling defects are reproduced below. Local targeted, full unit (2,954 tests), system (21 tests), RuboCop, and exact-head GitHub quality/security/validate/Docker checks are green; the findings come from focused adversarial failure injection and filesystem races. Requesting changes before merge.

AlexanderReaper7:
>[!NOTE]
>Written by opus 5

All seven findings from the second pass are fixed in c4bf2ff. Thanks — several of these were real holes in the recovery paths rather than in the happy path, and two of them changed the design rather than patching it.

**[P1] Abort when the publication journal cannot persist** — `record_publication!` no longer rescues. It is the only durable thing that can reverse bytes a move has already consumed, so a failed write now propagates into `import!`'s rollback, which reverses the in-memory record while it is still available. A journal that cannot be written can no longer produce an acquired book with `publication_record = nil`.

**[P1] Keep the publication journal until undo state commits** — `reverse_publication!` takes `clear_record:`, and `undo_import!` passes `false`. The journal is now cleared by the same transaction that resets the detection and releases the book, so a failure (or a process death) in that window leaves a record to retry from. Every reversal step was already idempotent, so the retry re-reverses only what is left; there is a test that forces `release_book_after_undo!` to fail and then completes the undo on the second call.

**[P1] Refuse undo when a moved source path has been reused** — `restore_moved_file!` now decides from the library artifact first (`:absent` means already reversed, `:mismatch` means hands off) and only then looks at the source. The importer records each file's source `(device, inode)` *before* publishing, so "the original is still there, this is a partly-reversed tree move" is now provable rather than inferred from `File.exist?`. Anything else occupying the path is a conflict: the occupant is left alone and the library keeps the only surviving copy.

**[P1] Restore moved files through the pinned watched-folder root** — the publication record now carries `source_base` (the canonical watched-folder root, from `WatchedFolderScanService.import_root`). Both the parent creation and the publication go through `FileCopyService.ensure_directory` / `mv_noreplace` rooted there, so the restore walks pinned `O_NOFOLLOW` descriptors from the root instead of trusting the recorded parent. A publication with no recorded root refuses to restore rather than falling back.

**[P2] Reconcile an interrupted durable publication before retrying** — `import!` now reconciles a journal left by a previous attempt before doing anything else. If that attempt also claimed the book, the import actually finished and only the bookkeeping was lost, so its result is handed back (no second publication, no numbered duplicate). Otherwise the stranded publication is reversed — releasing the dead attempt's reservation and returning a moved source — and the import runs from the restored source. Since a restore lands on a fresh inode, the detection's recorded identity is re-read so the retry's identity check still means something.

**[P2] Preserve the displayed match in one-click imports** — the queue's Import button submits `item.default_selection`, so it resolves exactly the candidate the row displays.

**[P2] Ignore failed jobs when restarting the scan chain** — `scan_job_pending?` now excludes `.where.missing(:failed_execution)`, matching `DownloadMonitorJob`. (No unit test: the Solid Queue tables live in a separate database that isn't prepared for the test suite here, the same reason the equivalent scopes elsewhere in the repo are untested. Worth a look in CI.)

Verification: full suite in the app image — 2961 runs, 11045 assertions, 0 failures, 0 errors. RuboCop `--fail-level warning` clean on the changed files. Six new tests cover the journal-write abort, the undo-transaction rollback and retry, the reoccupied source path, the swapped source ancestor, the interrupted-publication reversal, and the resumed claim. Brakeman, system tests, and mutant still need CI.

app/services/library_acquisition_service.rb
AlexanderReaper7: >[!NOTE]
>Written by opus 5

Fixed in c4bf2ff. `record_publication!` no longer rescues — the failure propagates into `import!`'s rollback, which reverses `importer.publication_record` while that in-memory record is still available.

Forcing the `update_columns` to fail now unwinds the move (source restored, library artifact removed, book un-acquired, reservation released) and re-raises, instead of reaching `claim_file_path!`. The rescue is also careful not to mask it: the reversal's own failure is logged rather than raised, and `refresh_source_identity!` was made non-raising for the same reason, since a reversed move puts the source back on a fresh inode and the next attempt has to be able to verify it.

Test: `import! fails instead of committing an import it could not journal`.

app/services/library_acquisition_service.rb
AlexanderReaper7: >[!NOTE]
>Written by opus 5

Fixed in c4bf2ff. `reverse_publication!` now takes `clear_record:`, and `undo_import!` passes `false` — the journal is nil'd by the same transaction that resets the detection and releases the book, so it is durable for exactly as long as there is state left to commit.

Forcing `release_book_after_undo!` to fail now rolls back to an `imported` detection **with its journal intact**, and the retry completes: every reversal step treats an already-removed file as `:absent`, so it re-reverses only what is left. Same for a process death in that window.

Test: `undo_import! keeps the journal until the state reset commits`, which asserts the rollback keeps the record and then runs the undo again to completion.

app/services/library_acquisition_service.rb
AlexanderReaper7: >[!NOTE]
>Written by opus 5

Fixed in c4bf2ff — and you're right that `File.exist?` was never evidence of anything.

`LibraryFileImporter#publish` now lstats the source **before** publishing and stores `source_device`/`source_inode` on the entry, so undo has something to compare against after a move has consumed it. `restore_moved_file!` decides from the library artifact first — `:absent` means this entry was already reversed, `:mismatch` means something else owns the path and neither end may be touched — and only then classifies the source:

- `:absent` → restore it
- `:match` → the original is still there (a tree move reversed before `consume_source_tree!`), so the library copy is the redundant one and gets discarded
- `:mismatch` → conflict: leave the occupant alone, keep the library artifact, return false so the undo raises

An entry that recorded no source identity counts as a mismatch rather than an assumed match, so it fails closed.

Your reproduction now leaves the unrelated bytes untouched, keeps the only surviving copy of the imported bytes in the library, and leaves the detection `imported`. Test: `undo_import! refuses to restore over a source path taken by a different file`.

app/services/library_acquisition_service.rb
AlexanderReaper7: >[!NOTE]
>Written by opus 5

Fixed in c4bf2ff. The publication record now carries `source_base` — the canonical watched-folder root, from a new `WatchedFolderScanService.import_root` — passed down from `DetectedImportJob` through `import!` into the importer.

The restore uses it for both steps: `FileCopyService.ensure_directory(File.dirname(source), root: source_base)` instead of `FileUtils.mkdir_p`, then `mv_noreplace(destination, source, root: source_base)`. Both walk from that root through pinned `O_NOFOLLOW` descriptors, so a descendant swapped for a symlink raises instead of redirecting the copy outside the watched folder. A publication with no recorded root refuses to restore rather than falling back to the recorded parent.

Test: `undo_import! will not restore through a swapped source ancestor symlink` — asserts nothing lands in the external directory and the library copy is kept rather than moved somewhere unreachable.

app/services/library_acquisition_service.rb
AlexanderReaper7: >[!NOTE]
>Written by opus 5

Fixed in c4bf2ff. `import!` now runs `recover_interrupted_publication!` before it reserves anything, so an owner that already carries a journal is reconciled rather than re-imported.

Both branches you allowed for are implemented:

- **Resume.** If the earlier attempt also claimed the book — `publication_claimed_by?` matches `book.file_path` against the published file, or against the per-book directory everything was published into — the import actually finished and only the caller's bookkeeping was lost. Its result is handed straight back, so no second publication and no numbered duplicate. The library scan is re-triggered, since the dead attempt may have died before announcing it.
- **Reverse.** Otherwise the stranded publication is reversed, after releasing that attempt's reservation (which would otherwise fail `reserve_book!` — `release_reservation!` only touches a reservation still held by this owner on an unclaimed book, i.e. exactly this case). The restored source's identity is re-read, because the restore lands on a fresh inode and the detection's recorded pair would otherwise fail the importer's own check. If the reversal can't complete, it raises `AcquisitionConflictError` naming the output path instead of importing from a source that isn't there.

This also covers the case where the *previous* attempt's rollback failed: that publication record is deliberately kept, and the next attempt retries the reversal.

Tests: `import! reverses an interrupted publication instead of re-importing a consumed source` (worker killed via a non-`StandardError`, so the rollback genuinely never runs) and `import! resumes an interrupted attempt that already claimed the book`.

app/views/admin/detected_imports/index.html.erb
AlexanderReaper7: >[!NOTE]
>Written by opus 5

Fixed in c4bf2ff. The button now submits `params: { selection: item.default_selection }` — the same value the review form pre-selects — so `apply_selection!` resolves the displayed candidate (`work:` or `book:`) instead of leaving `suggested_book` nil and letting the job invent a book from the filename parse.

A candidate that can no longer be resolved still raises `SelectionError` and redirects to Review with the reason, rather than importing something else.

Test: `the queue's one-click Import submits the match the row displays` — asserts the rendered hidden input carries `book:<id>`, and that posting it attaches that book rather than the deliberately wrong parsed title.

app/jobs/watched_folder_scan_job.rb
AlexanderReaper7: >[!NOTE]
>Written by opus 5

Fixed in c4bf2ff — `scan_job_pending?` now adds `.where.missing(:failed_execution)` and widens its rescue to `ActiveRecord::ActiveRecordError`, matching `DownloadMonitorJob#monitor_job_pending?`.

Flagging one gap honestly: this is the only one of the seven with no test. The `solid_queue_*` tables live in the `queue` database, which `db:test:prepare` doesn't create in this environment, and the predicate also short-circuits on `solid_queue_adapter?` under the `:test` adapter — which is why the equivalent scopes on `DownloadMonitorJob`, `UploadRecoveryJob` and friends are untested too. My first attempt at a test turned into a silent skip because the predicate rescues the missing-table error into `false`, so I removed it rather than leave something that looks like coverage. Worth confirming in CI if you have the queue schema loaded there.

AlexanderReaper7:
Pushed `865159a` — follow-up fixes plus a deduplication pass.

**Behaviour**

- Background scans now record running/completed status too. Previously only manual scans did, so the queue would offer "Scan now" during a background scan — the concurrency key then discarded the manual job after the admin had been told it started.
- Removing a detection whose source is still in the watched folder is refused. The row is the only thing suppressing re-detection, so removing it resurrected the file (with a fresh notification) on the next scan. Removal stays available once the source is gone.
- A detection whose source moved to another path is retired instead of sitting in the queue beside the row that replaces it.
- An attempt killed between `reserve_book!` and the journal write now releases its reservation. Nothing else clears one held by that owner, so it blocked the book for the upload and download pipelines permanently.

**Deduplication**

- *"Stuck importing"* was defined three times — a model predicate, raw SQL in `DetectedImportJob#claim`, and a where-chain in the controller — and had already drifted on `NULL updated_at`. Now one `.stuck_importing` scope plus `.claimable`, with the Ruby predicate kept as its in-memory mirror (the queue asks it per rendered row).
- The *lstat-and-compare-identity* check existed in three copies under three vocabularies (`:live/:missing/:replaced`, `:match/:absent/:mismatch`, and a boolean). Now `FileCopyService.path_identity_state`, returning `:absent` / `:match` / `:mismatch`, with unprovable reported as `:mismatch` so no caller consumes a path it cannot verify.
- `output_base_path` was a fourth copy of the output-root lookup and disagreed with `PathTemplateService` on the fallback branch. Promoted to `PathTemplateService.output_root_for` (+ `.output_roots`), used by both the acquisition service and the scanner, so the root an import publishes under and the root its templates render beneath cannot diverge. The copies in `PostProcessingJob` and `UploadImportFileService` are left alone to keep this PR scoped.
- `Book::RESERVATION_CLEARED` plus `.reserved_by` / `.unclaimed` scopes replace the repeated reservation triple and its where-chain.

Comments were also trimmed throughout — the load-bearing reasoning stays, the sentences restating the code below them don't.

**Verification:** 515 tests across the feature files and every consumer of the shared code touched (`file_copy_service`, `upload_import_file_service`, `direct_download_file_service`, `safe_library_deletion_service`, `upload_processing_job`, `book`, `path_template_service`) — 1947 assertions, 0 failures. Rubocop clean on all changed files.

Pedro-Revez-Silva:
Closing as stale: no activity since 31 Jul (over two weeks). Please reopen if you are still working through the review comments.

PR #411, opened 2026-07-29
fix: support all-squash NFS ownership for non-root PUIDs

### LINKED ISSUE #409: Post-processing failed: A safe filesystem operation failed (NFS / Synology Permission Mismatch)
I am encountering a `Post-processing failed: A safe filesystem operation failed` error when Shelfarr attempts to process downloaded books.

## Environment
* Host OS: Ubuntu Server
* Storage: Synology NAS connected via NFS v4
* App Deployment: Docker container running on the Ubuntu server
* Mount options (/etc/fstab): `XX.XX.XX.XX:/volume2/data/ /mnt/data nfs defaults 0 0`

## Technical Details & Troubleshooting Done
My Synology NFS share is currently configured with the Squash option set to "Map all users to admin".
When checking the permissions of the downloaded book folder from the Ubuntu server (`ls -ln /mnt/data/downloads/books`), the files show up as owned by the standard Unix `nobody` user (`65534`):

```
total 20
drwxrwxrwx 2 65534 100 4096 Jul 29 04:50 'Leigh Bardugo - Ruin and Rising'
```

Because Synology maps these actions to `nobody` or masks them under the admin user, Shelfarr's strict POSIX safety checks (advisory locks/flock, hard links, or chmod operations) fail during post-processing.
Changing the NFS Squash setting to "**No mapping**" on the NAS resolves the issue for Shelfarr, but it alters file ownership behavior for other automated applications in my existing Docker stack that rely on the current mapping setup. Setting Shelfarr's `PUID` to `65534` is also blocked by design.

## Questions / Desired Outcome
   1. Is there a way to bypass or relax this specific filesystem safety check via an environment variable for environments utilizing NFS squash mappings?
   2. Alternatively, what is the officially recommended approach for handling Shelfarr post-processing when the underlying network storage forces user squash mappings?

MathiasOki:
## Summary
- generalize the ownership-squash detection probe in `private_entry_owned_by_process?` to run whenever an entry's uid doesn't match our own, not only when running as root
- this covers NFS mounts that squash *all* uids (e.g. Synology's "map all users" mode), where a non-root PUID container's own uid gets remapped too, so the old root-only gate rejected every staging/lock/quarantine entry as unsafe
- add regression coverage for a non-root process under all-uid squash
- fix two existing tests that simulated "different owner" only by stubbing `Process.euid` without an actual differing owner, which the more thorough probe now passes vacuously; they use a real `File.chown` (gated on running as root) instead

## Test plan
- `bin/rails test test/services/file_copy_service_test.rb` (138 runs, 0 failures, 4 skips — skips are root-gated cases not exercised in this non-root environment)
- `bin/rails test` full suite — no regressions; the only failures/errors present reproduce identically on `main` (missing `libvips` locally, and one pre-existing unrelated flaky test)

Closes #409


app/services/file_copy_service.rb
```diff
@@ -1347,12 +1347,13 @@ def mv(src, dest)
 
     private
 
-    # root_squash changes the UID reported for root-created entries. Discover
-    # that mapping from an exclusive entry on the same filesystem.
+    # NFS root_squash/all_squash remaps every uid to a fixed value (0 for
+    # classic root_squash, or e.g. 65534 for Synology "map all users").
+    # Discover that mapping from an exclusive entry on the same filesystem
+    # rather than assuming only root ever gets squashed.
```
Pedro-Revez-Silva: [P1] Do not treat an all-squash UID match as proof of application ownership. NFS `all_squash` maps every client UID/GID to the same anonymous identity, so an entry pre-created by another client and this fresh probe necessarily have the same `st_uid` and `st_dev`. I reproduced this with a pre-created private lock: this head accepts it and enters the critical section for a simulated non-root mapped UID, while `main` raises `UnsafePathError`. The same check protects private staging parents/files, and modes `0600`/`0700` cannot isolate clients once the server sees them as the same owner. Please keep this boundary fail-closed or require an explicit trusted-share opt-in/design that does not claim the probe identifies the creating process.

Pedro-Revez-Silva (changes requested):
Requesting changes for one verified security-boundary regression. I reproduced this against the exact head commit; the focused suite otherwise passes.

app/services/file_copy_service.rb
MathiasOki: You're right that an all-squash match doesn't prove ownership the way a root-squash match does, since every client collides into the same identity, not just root. Fixed in 9b20d1e: restored the fail-closed default for non-root processes (`return false unless effective_uid.zero?`), and gated the non-root probe path behind an explicit `TRUST_NFS_UID_SQUASH` env var that an operator must set, documented as an assertion that the export is single-tenant rather than proof the probe identifies the creating process. Added a regression test confirming a non-root process under all-squash still raises `UnsafePathError` without that opt-in, plus a test for the opt-in path.

Pedro-Revez-Silva (approved):
Reviewed head cf69a13d1cbd8951a618328befa4d01e280e133c against merge-base fdd8dd05f8c984b8980f70e99525999b8497d151. The earlier all-squash trust-boundary issue is fixed: non-root mismatched ownership remains fail-closed unless the operator explicitly asserts a single-tenant share with TRUST_NFS_UID_SQUASH=true. Local verification passed the focused tests, the 139-test FileCopyService suite, the full 2,894-test Rails suite, and changed-file quality checks. Exact-head GitHub CI, security, companion validation/smoke, and both image builds are green.

PR #446, opened 2026-08-14
Honor TZ for application timestamps

brandon-dacrib:
## What changed

- Configure Rails' application time zone from the standard `TZ` environment variable.
- Retain UTC as the default and keep Active Record timestamp storage in UTC.
- Pass `TZ` to both services in the Compose example.
- Document the IANA time-zone requirement and add a boot-level regression test.
- Keep the child boot probe out of the parent SimpleCov run so it cannot corrupt or pollute coverage results.

## Why

Setting `TZ` changed the container's libc clock, but Rails continued to render timestamps in UTC. After this change, an IANA zone such as `America/New_York` or `Europe/Lisbon` makes Rails-rendered timestamps follow the configured zone while database values remain UTC.

## Maintainer review

- Rebased onto current `main` with Brandon's original authorship preserved.
- Verified the default boot reports `UTC|utc` for Rails and Active Record.
- Verified `TZ=Europe/Lisbon` reports `Europe/Lisbon|WEST|utc`.
- Verified the Compose example parses successfully.
- Repaired the boot test so a parent `COVERAGE=1` is not inherited by the nested Rails process.

## Verification

- `bin/quality push` — passed after the coverage-process repair
- Focused boot regression — 1 run / 2 assertions, 0 failures
- Focused regression under `COVERAGE=1` — passed
- RuboCop — no offenses
- `git diff --check` — passed


Pedro-Revez-Silva:
Thanks @brandon-dacrib — I’ve adopted this pull request for maintainer integration. I’ll handle the rebase, any required adjustments, and final verification directly on this branch while preserving your commits and attribution. No action is required from you; I’ll only reach out if I need input on a product decision.

PR #447, opened 2026-08-14
Feat/per user library routing

### LINKED ISSUE #445: Library routing per user
I would like to see a feature that can route a requested item (ebook or audiobook) to a Library on a per user basis.  it would be nice to have a feature where you can set pathing for a user's request to their specific library once the download is completed.

rodrap50:
## Summary

Closes #445

- Adds three nullable columns to `users` (`preferred_audiobook_library_id`, `preferred_ebook_library_id`, `preferred_comicbook_library_id`)
- When a download completes, `PostProcessingJob` and `DownloadJob` check the requesting user's preferred library ID and fall back to the system default if unset
- Admins configure routing per user via a dedicated **Library Routing** page linked from the Actions column in Admin → Manage Users

## Test plan

- [ ] Run `bin/rails db:migrate` and verify the three new columns exist on `users`
- [ ] As admin, open Manage Users and confirm the **Library Routing** link appears in the Actions column for each user
- [ ] Set a library ID for a user and save — confirm it persists
- [ ] Trigger a download as that user and confirm the library scan targets the overridden ID
- [ ] Leave all fields blank for a user — confirm downloads fall back to the system-wide library setting

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,8 @@
 # See https://git-scm.com/docs/gitattributes for more about git attribute files.
 
+# Normalize line endings to LF for all text files (prevents CRLF in containers)
+* text=auto eol=lf
+
 # Mark the database schema as having been generated.
 db/schema.rb linguist-generated
 
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -1,6 +1,6 @@
 module Admin
   class UsersController < BaseController
-    before_action :set_user, only: [:show, :edit, :update, :destroy]
+    before_action :set_user, only: [:show, :edit, :update, :destroy, :library_routing, :update_library_routing]
 
     def index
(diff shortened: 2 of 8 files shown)


Pedro-Revez-Silva:
Thanks @rodrap50 for working on this. I’m closing this implementation because it does not address the core requirement in #445: per-user filesystem pathing. It changes the Audiobookshelf library scan target, but completed files still use the global library paths, so users are not actually routed to separate libraries. The current synchronization/cache flow would also need a broader per-user design. That is a fundamental scope change rather than a small integration fix, so I’m leaving #445 open for a fresh implementation instead of rewriting this PR in place. Your contribution and the design exploration remain credited here.

rodrap50:
Thanks.  I will restart fresh and work on this again, unless someone else picks it up before I get to it.

Issue #54, opened 2026-01-04
Feature Request: More Granular Status


human-cowculator:
**Description**  
When submitting a request, it appears to remain in a **Pending** state, and it is not clear what the system is waiting on or which step of the process it is currently in. It is possible this is already tracked internally, but that information does not seem to be exposed in the UI.

**Suggested Enhancements**
- More granular request statuses after submission or approval, for example:
  - Awaiting approval (if applicable)
  - Searching
  - Downloading (with progress, if available)
  - Importing
  - Completed / Failed
- Optional **Approve / Deny** controls for requests, if approval is part of the workflow
- Clear indication of which step failed when a request does not complete
- A **Retry Search** (or similar) action when a request fails, without needing to recreate it

**Why This Would Help**
- Makes it easier to understand whether a request is idle, waiting, or actively processing
- Reduces confusion around requests that appear “stuck” in Pending
- Improves transparency without requiring log inspection

PR #452, opened 2026-08-18
fix: resolve download and filesystem backlog

### LINKED ISSUE #401: WSL2/NTFS post-processing fails during quarantine cleanup
Post-processing fails with RuntimeError at "Copying library content" step

Description:

Post-processing consistently fails at the "Copying library content" step with a RuntimeError that is swallowed with no additional detail logged. This affects all file formats (epub, azw3, m4b tested).

Environment:

Shelfarr running in Docker on WSL2 (Windows)
qBittorrent running on Windows host
Remote path mapping enabled: D:\QbittorrentMove → /downloads
Ebook destination: /ebooks

Behavior:

Download is added to qBittorrent successfully 
Download monitor detects completion correctly 
Path resolves via global_prefix_remap 
Destination folder is created correctly 
File copy fails with RuntimeError 

Logs:
docker logs shelfarr 2>&1 -f | grep -i "380\|RuntimeError\|copy\|staging"
{"time":"2026-07-27T15:12:16.380296384-06:00","level":"INFO","msg":"Request","path":"/requests","status":200,"dur":1928,"method":"GET","req_content_length":0,"req_content_type":"","resp_content_length":7512,"resp_content_type":"text/html; charset=utf-8","remote_addr":"2a02:6ea0:d704:6255::26","user_agent":"Mozilla/5.0 (Android 17; Mobile; rv:152.0) Gecko/152.0 Firefox/152.0","cache":"miss","query":"status=not_found","proto":"HTTP/1.1"}
{"time":"2026-07-27T15:12:28.380249098-06:00","level":"INFO","msg":"Request","path":"/requests","status":200,"dur":3734,"method":"GET","req_content_length":0,"req_content_type":"","resp_content_length":236217,"resp_content_type":"text/html; charset=utf-8","remote_addr":"2a02:6ea0:d704:6255::26","user_agent":"Mozilla/5.0 (Android 17; Mobile; rv:152.0) Gecko/152.0 Firefox/152.0","cache":"miss","query":"attention=true","proto":"HTTP/1.1"}
[61a64b91-1dc0-401e-a097-569f98dc7235] [ActiveJob] Enqueued DownloadJob (Job ID: baf05af8-fed9-44f4-815a-d8632bd7ed6c) to SolidQueue(default) with arguments: 380
[ActiveJob] [DownloadJob] [baf05af8-fed9-44f4-815a-d8632bd7ed6c] Performing DownloadJob (Job ID: baf05af8-fed9-44f4-815a-d8632bd7ed6c) from SolidQueue(default) enqueued at 2026-07-27T21:13:18.388780521Z with arguments:380
[ActiveJob] [DownloadJob] [baf05af8-fed9-44f4-815a-d8632bd7ed6c] [DownloadJob] Starting download #380 for request #407
[ActiveJob] [DownloadJob] [baf05af8-fed9-44f4-815a-d8632bd7ed6c] [DownloadJob] Using client 'qbitorrent' for download #380
[ActiveJob] [DownloadJob] [baf05af8-fed9-44f4-815a-d8632bd7ed6c] [DownloadJob] Successfully added torrent for download #380, external_id: 986d46cca40d2a938986df871602906698824b00
[ActiveJob] [DownloadMonitorJob] [43fa11a0-a0a4-4099-b203-0381569ee9ed] [DownloadMonitorJob] Download 380 completed
[ActiveJob] [DownloadMonitorJob] [43fa11a0-a0a4-4099-b203-0381569ee9ed] Enqueued PostProcessingJob (Job ID: fb20bad9-aec5-47de-82e1-1adf7b108009) to SolidQueue(default) with arguments: 380
[ActiveJob] [PostProcessingJob] [fb20bad9-aec5-47de-82e1-1adf7b108009] Performing PostProcessingJob (Job ID: fb20bad9-aec5-47de-82e1-1adf7b108009) from SolidQueue(default) enqueued at 2026-07-27T21:14:24.475706743Z with arguments: 380
[ActiveJob] [PostProcessingJob] [fb20bad9-aec5-47de-82e1-1adf7b108009] [PostProcessingJob] Starting download #380, book #406
[ActiveJob] [PostProcessingJob] [fb20bad9-aec5-47de-82e1-1adf7b108009] [PostProcessingJob] Copying library content
[ActiveJob] [PostProcessingJob] [fb20bad9-aec5-47de-82e1-1adf7b108009] [PostProcessingJob] Download #380 failed: RuntimeError
[ActiveJob] [OwnedLibraryAutomationJob] [8d18a55f-90cf-43cb-8c06-d81b12016d02] Performing OwnedLibraryAutomationJob (Job ID: 8d18a55f-90cf-43cb-8c06-d81b12016d02) from SolidQueue(default) enqueued at 2026-07-27T21:15:22.354127380Z
[2b896620-be36-42aa-9a93-a07cea13f606] [ActiveJob] Enqueued PostProcessingJob (Job ID: c7954c20-df76-40d8-8067-1eb2b9b2816a) to SolidQueue(default) with arguments: 380, 0, "fb20bad9-aec5-47de-82e1-1adf7b108009"
[ActiveJob] [PostProcessingJob] [c7954c20-df76-40d8-8067-1eb2b9b2816a] Performing PostProcessingJob (Job ID: c7954c20-df76-40d8-8067-1eb2b9b2816a) from SolidQueue(default) enqueued at 2026-07-27T21:27:11.414141763Z with arguments: 380, 0, "fb20bad9-aec5-47de-82e1-1adf7b108009"
[ActiveJob] [PostProcessingJob] [c7954c20-df76-40d8-8067-1eb2b9b2816a] [PostProcessingJob] Starting download #380, book #406
[ActiveJob] [PostProcessingJob] [c7954c20-df76-40d8-8067-1eb2b9b2816a] [PostProcessingJob] Copying library content
[ActiveJob] [PostProcessingJob] [c7954c20-df76-40d8-8067-1eb2b9b2816a] [PostProcessingJob] Download #380 failed: RuntimeError
[39df010c-8274-4e59-82d3-9fb40829630b] Completed 200 OK in 3763ms (Views: 380.1ms | ActiveRecord: 177.8ms (1625 queries, 1508 cached) | GC: 495.7ms)
{"time":"2026-07-27T15:27:21.621654568-06:00","level":"INFO","msg":"Request","path":"/requests","status":200,"dur":3800,"method":"GET","req_content_length":0,"req_content_type":"","resp_content_length":230370,"resp_content_type":"text/html; charset=utf-8","remote_addr":"95.173.221.177","user_agent":"Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/150.0.0.0 Safari/537.36","cache":"miss","query":"","proto":"HTTP/1.1"}

Additional info:

Manually copying the file inside the container works fine: cp "/downloads/Evil Roots_ Killer Tales of the - Daisy Butcher.azw3" "/ebooks/Daisy Butcher - Evil Roots/"
The destination folder /ebooks is writable by the rails user
This started after the recent update that fixed AtomicPublicationUnsupportedError and the symbolic link safety check errors
The RuntimeError is being caught and swallowed with no stack trace or inner exception logged, making it impossible to diagnose from outside

Possible cause:
The fix for the previous AtomicPublicationUnsupportedError may have introduced a new code path that fails on cross-filesystem copies (WSL /downloads → Windows NTFS /ebooks) without proper error handling.

### LINKED ISSUE #429: Direct-download staging is not safely accessible: filesystem did not preserve safe directory permissions
Hello,

I'm receiving this error on every book download:
`Direct-download staging is not safely accessible: filesystem did not preserve safe directory permissions`

I save the ebooks to a Synology NAS mounted on a server.
This is the /etc/fstab line:

`//XXX.XXX.XXX.XXX/YAMS /srv/media nfs auto,defaults,nofail 0 0`

DiskStation has the correct permission,

This problem started appearing a few weeks ago, could you please guide me to solve this problem?

Thank you!

### LINKED ISSUE #437: Comic Vine issue metadata is stored correctly but issue number is not effectively used for Prowlarr search/matching
Hi! I wanted to report something I found while testing the newer Comics & Manga / Comic Vine workflow.

A quick disclaimer: I am not a developer. This was found while troubleshooting my Shelfarr install with AI assistance, so I may be misunderstanding some of the intended behavior. I wanted to provide the details because the reproduction seems fairly clean and I hope it is useful.

## Summary

Comic Vine appears to be supplying the correct issue-level metadata, and Shelfarr stores that metadata correctly.

However, when Shelfarr searches Prowlarr for the issue, the search does not appear to make effective use of the stored issue number / series position.

In my test this caused:

1. The initial Prowlarr search to return no results.
2. Shelfarr to fall back to broader/generic searching.
3. Shelfarr to eventually search without category restrictions.
4. 570 Prowlarr results to be returned.
5. None of the visible candidates to be an obvious exact match for the requested issue.
6. The highest confidence result to be only 47%.

Auto Select was disabled, so Shelfarr correctly stopped for manual review and did not download anything.

## Test case

Comic Vine volume:

- Series: `Batman`
- Comic Vine volume ID: `4050-796`

I selected only issue  # 1:

- Comic Vine issue ID: `4000-105811`
- Issue: `1`
- Release date: `1940-04-01`

Only one issue was selected/requested.

The request creation logs correctly showed:

```text
content_kind=>"graphic"
book_types=>["comicbook"]
collection_source=>"comic_vine"
collection_id=>"4050-796"
collection_item_ids=>["comic_vine:4000-105811"]
````

Shelfarr also reported:

```text
[CollectionRequestExpansionJob] created 1 requests for comic_vine:4050-796
```

So this was not accidentally a whole-volume request.

## Stored database metadata

I checked the resulting request/book record read-only with SQLite.

For request Num5, Shelfarr stored:

```text
5|Batman - #1 - The Legend of the Batman: Who He Is and How He Came To Be; The Joker||1|Batman|1|1940-04-01
```

The fields queried were:

```text
request id
title
author
issue_number
series
series_position
release_date
```

So the relevant metadata is present:

```text
title:
Batman - #1 - The Legend of the Batman: Who He Is and How He Came To Be; The Joker

issue_number:
1

series:
Batman

series_position:
1

release_date:
1940-04-01
```

## Acquisition behavior

When the request queue processed it:

```text
[RequestQueueJob] Processing 1 pending requests (batch_size: 5)
[RequestQueueJob] Enqueued SearchJob for request #5
[SearchJob] Starting search for request #5
```

The first Prowlarr attempt then failed to find anything:

```text
[SearchJob] Prowlarr book search returned no results for request #5; retrying with a generic query
```

It later broadened the search:

```text
[SearchJob] Supplementing Prowlarr search for request #5 without category restrictions
[SearchJob] Found 570 Prowlarr results
[SearchJob] Total 570 results for request #5
[SearchJob] Auto-select disabled, flagged for manual selection for request #5
```

The resulting candidate list contained many loosely related Batman results, but I could not identify an exact match for the requested Batman # 1. The highest confidence result was only 47%.

I did not select or download that result.

## What I expected

For a `comicbook` request where Shelfarr already knows:

```text
series = Batman
issue_number = 1
series_position = 1
release_date = 1940-04-01
```

I expected comic-specific searching/matching to make use of at least the series + issue number.

For example, perhaps issue-aware query variants such as:

```text
Batman 1
Batman 001
Batman #1
```

or whatever approach makes the most sense for the different naming conventions used by comic releases/indexers.

I would also expect the release confidence scoring for a comicbook request to strongly reward an exact issue-number match and penalize a conflicting issue number.

## Possible cause

I looked through the current code with AI assistance.

From what I can tell, the Prowlarr search path in `SearchJob` primarily works from the stored title/author/book type and generic title queries.

I also could not see `issue_number` or `series_position` being used as a scoring factor in `ReleaseScorer`; its scoring appears to be based on title, author, language, format, and availability.

Please take that interpretation with a grain of salt since I am not a Ruby developer.

The important part I can confirm from testing is that Comic Vine's issue metadata reaches Shelfarr and is stored correctly, but that specificity seems to be lost during acquisition searching/matching.

## Environment

* Shelfarr Docker image: `ghcr.io/pedro-revez-silva/shelfarr:latest`
* Docker Engine: 29.7.2
* Docker Compose: v5.4.0
* Debian 13
* Indexer provider: Prowlarr
* Search Scope: Broad
* Auto Select: OFF
* Auto Approve Requests: OFF
* Minimum Match Confidence: 50
* Comic Vine API connection test: PASS
* Metadata Source: Auto

Metadata provider priority during the final test:

```text
hardcover,comic_vine,openlibrary,google_books
```

## Additional Comic Vine validation

Before creating the request I also tested Comic Vine directly inside Shelfarr:

* Comic Vine connection test passed.
* Forcing Metadata Source to Comic Vine returned Comic Vine search results.
* Returning Metadata Source to Auto also returned Comic Vine results after adjusting provider priority.
* Comic Vine supplied the correct volume and individual issue list.
* Shelfarr created exactly one `comicbook` request when only Batman # 1 was selected.

So this does not appear to be a Comic Vine API/connectivity problem.

If there are any additional logs, database fields, searches, or controlled tests that would help, I am happy to run them.

Thanks for the work on Shelfarr. I appreciate the time and effort you have put into this project and I just wanted to contribute here in case it helps.

### LINKED ISSUE #441: `reference` mode can't import from a symlink source at all — breaks the standard debrid downloads-folder pattern (decypharr/Zurg/blackhole-style setups)
`reference` mode (#415) is meant to let a library entry be a real symlink pointing at wherever the download lives, avoiding double storage for debrid/rclone-mounted content. In practice, it can't accept a symlink source at all: `with_pinned_regular_child` — used by both `with_pinned_source` (reference) and `with_pinned_hardlink_source` (hardlink) — opens the source with `O_NOFOLLOW` unconditionally. If the immediate source is itself a symlink, it raises `ELOOP` (`FileCopyService::UnsafePathError`), regardless of import mode.

Confirmed directly: symlinking a file into a downloads folder pointing at a real (rclone-mounted) target, then calling `FileCopyService.reference_noreplace` against that symlink, fails with:

```
FileCopyService::UnsafePathError: source path contains a symbolic link or non-regular path: Too many levels of symbolic links - openat
```

This means `reference` mode only works when the download client reports a path to a real file *directly* — it fails against the standard way most debrid tools present content: an intermediate symlink staged in a downloads folder, pointing at the real mount. That's how decypharr, Zurg, blackhole-style setups, and torbarr's own default (non-passthrough) mode all work — presumably the primary use case #415 was built for.

**Request:** let `reference` mode resolve one level of symlink indirection when the immediate source is itself a symlink — read the link, verify the target is a real regular file, and point the new library symlink at *that* final target instead of refusing outright. This fixes the core problem (reference mode currently can't be used against the pattern it's supposedly for) and, as a direct side effect, makes the downloads-folder-cleanup problem solvable too: once the library symlink no longer depends on the downloads-folder path, deleting that local copy afterward becomes safe — a plain `rm` of the path Shelfarr already has recorded via `download.download_path`, not a call to the download client. That sidesteps #427's blocker entirely, since it never touches remote/shared torrent state, only a directory this one `Download` row already owns.

Two other shapes worth considering, not mutually exclusive with the above:

1. Let `hardlink` mode hardlink a symlink *object* when the immediate source is itself a symlink (an `O_NOFOLLOW`-safe `linkat` of the link, with readlink-based identity verification instead of the regular-file manifest check it uses today) — opt-in, not the current silent refusal.
2. Give `move` mode a real `rename()`-based path when source and destination share a filesystem, which for a symlink source naturally relocates it instead of copying through it; fall back to today's copy+delete only across filesystem boundaries.

For context: Sonarr/Radarr use a remote-delete approach (`removeCompletedDownloads` calling the download client's own delete API) against this same torbarr backend, and it works reliably in practice. Not meant as a rebuttal to #427's findings — Sonarr/Radarr's implementations are mature and almost certainly have their own request/queue-item identity tracking built up over years that avoids the exact cross-request race described there. It's just supporting evidence that the pattern is sound when the ownership tracking behind it is solid, which is arguably an argument for the local-only delete above being the lower-risk path for Shelfarr specifically, since it sidesteps needing that ownership machinery at all.

Not a duplicate of #427 — that's about calling the download client's *remote* delete API after import, which you found unsafe due to cross-request torrent-ownership races. The core ask here (resolving one level of symlink indirection) is purely local filesystem handling in `FileCopyService`, and the cleanup that falls out of it never touches the download client's remote state.

(Current workaround on our end: configuring the download client to report the rclone mount path directly instead of an intermediate downloads-folder symlink, so `reference` mode's existing "create a new symlink" behavior gets a real file to work with. Works, but it's only possible because we control the download client's own source and could add that reporting mode ourselves — not something available to anyone using an off-the-shelf debrid tool that stages content the standard way.)

### LINKED ISSUE #444: Shelfarr does not appear to respect Decypharr’s download action settings
I'm using Shelfarr with Decypharr and Torbox.

Decypharr has its download action configured as **symlink**.

When a torrent is submitted by Prowlarr to the same Decypharr instance, Decypharr correctly processes it as:

[Torbox] Processing torrent Action=symlink Arr=prowlarr

However, when the same type of torrent is submitted by Shelfarr, Decypharr processes it as:

[Torbox] Processing torrent Action=download Arr=shelfarr

It then downloads the files from Torbox instead of using the configured **symlink** action.

For example:

[manager] Download completed, processing action action=download
[manager] Downloading 39 files...
[manager] Downloaded all files for ...

Prowlarr is able to use the same Decypharr instance and correctly gets the configured symlink behaviour.

I've also checked Shelfarr's qBittorrent/Decypharr adapter and could not find an obvious setting that should be forcing `Action=download`.

Is Shelfarr sending something in its request to Decypharr that causes Decypharr to use `Action=download`, rather than respecting Decypharr's configured **symlink** download action?

I'm trying to understand why Shelfarr appears to override or bypass Decypharr's configured download action.

Pedro-Revez-Silva:
## Summary

- respect Decypharr download-action configuration instead of forcing direct downloads
- search and score Comic Vine releases by strict series, issue, run-year, language, and format identity
- backfill missing Comic Vine run-year metadata and fail safe on explicit years until it is available
- label comics consistently as Comics & Manga across dashboard, request, and library surfaces
- support descriptor-pinned reference imports from top-level links and directories of symlink leaves
- persist immutable reference target boundaries for later download and deletion authorization
- move new direct downloads to a private v2 staging namespace without touching unsafe legacy state
- add bounded, serialized DrvFS publication and private-staging cleanup while preserving no-clobber guarantees
- reserve all internal staging namespaces across templates, uploads, imports, and deletion
- keep the Libation companion container smoke test portable across GNU/Linux and macOS date implementations
- update json, mail, net-imap, and sqlite3 to patched versions required by the security gate

## Test plan

- [x] Added regression tests that failed against main before each fix
- [x] Focused download/scoring regression suite: 38 tests, 230 assertions
- [x] Focused comic-label rendering suite: 187 tests, 883 assertions
- [x] Full bin/quality push in Ruby 3.3.6 with Chromium
- [x] Full Rails and system suites
- [x] Coverage: 91.19% line, 73.28% branch
- [x] bundler-audit --update: no vulnerabilities
- [x] CIFS serverino and noserverino filesystem contracts
- [x] Libation companion tests and container smoke test
- [x] Manual local UI validation: login, dashboard, request results, comic run-year scoring, completed request detail, and authorized EPUB download
- [x] Manual comic-label UI validation: dashboard recent items and requests, requests list, library grid, and library details
- [x] Browser console and page logs clean during the validated flows
- [x] Repeated adversarial review of the exact PR head with no remaining verified findings
- [x] All GitHub Actions and CodeQL checks green on e70eb5c491cbdb841f9f72e004cf98aef76c5513

Closes #444
Closes #441
Closes #437
Closes #429
Closes #401

### DIFF
--- a/app/jobs/post_processing_job.rb
+++ b/app/jobs/post_processing_job.rb
@@ -291,7 +291,7 @@ class PostProcessingJob < ApplicationJob
     when FileCopyService::DurabilityUnsupportedError
       "The library filesystem cannot safely complete destructive imports because fsync is unsupported; use another filesystem or retain the download source"
     when FileCopyService::AmbiguousPublicationError
-      "An incomplete library publication was retained for manual review; preserve the destination and .shelfarr-copy artifacts before retrying"
+      "An incomplete library publication was retained for manual review; preserve the destination and adjacent Shelfarr filesystem artifacts before retrying"
     when FileCopyService::AtomicPublicationUnsupportedError
       if bounded_exception_message(error).match?(/manual (?:cleanup|review)/i)
         "Interrupted library publication artifacts were retained for manual review before retrying"
--- a/app/jobs/post_processing_job.rb
+++ b/app/jobs/post_processing_job.rb
@@ -277,6 +277,14 @@ class PostProcessingJob < ApplicationJob
       location ? "#{message} at #{location}" : message
     when FileCopyService::DurabilityUnsupportedError
       "The library filesystem cannot safely complete destructive imports because fsync is unsupported; use another filesystem or retain the download source"
+    when FileCopyService::AmbiguousPublicationError
+      "An incomplete library publication was retained for manual review; preserve the destination and .shelfarr-copy artifacts before retrying"
+    when FileCopyService::AtomicPublicationUnsupportedError
+      if bounded_exception_message(error).match?(/manual (?:cleanup|review)/i)
+        "Interrupted library publication artifacts were retained for manual review before retrying"
+      else
+        safe_attention_detail(error)
+      end
     when FileCopyService::UnsafePathError
       "The download contains a symbolic link or non-regular path, or changed during its safety checks"
     when AudiobookBundleImportPlanner::UnsafeDestinationError
--- a/app/jobs/post_processing_job.rb
+++ b/app/jobs/post_processing_job.rb
@@ -39,6 +39,7 @@ class PostProcessingJob < ApplicationJob
     @hardlink_fallback_copied_count = 0
     @referenced_file_count = 0
     @reused_file_count = 0
+    @completed_reference_target_roots = []
     download = Download.find_by(id: download_id)
     return unless download&.completed?
 
@@ -96,11 +97,15 @@ class PostProcessingJob < ApplicationJob
       if source_path_unavailable?(source_path)
         return retry_source_path_later(download, request, source_path, source_path_retry_count)
       end
-      source_authorization = validate_download_specific_source_path!(
-        source_path,
-        source_resolution[:authorized_roots],
-        download
-      )
+      source_authorization = begin
+        validate_download_specific_source_path!(
+          source_path,
+          source_resolution[:authorized_roots],
+          download
+        )
+      rescue FileCopyService::ReferenceTargetUnavailableError
+        return retry_source_path_later(download, request, source_path, source_path_retry_count)
+      end
 
       # Reference mode keeps download bytes as the only content; never delete
       # usenet/client sources after a symlink-only import.
@@ -189,15 +194,23 @@ class PostProcessingJob < ApplicationJob
       end
 
       if book.file_path.blank?
+        reference_target_roots = Book.dump_reference_target_roots(
+          @completed_reference_target_roots
+        )
         claimed = Book.where(id: book.id)
           .where("file_path IS NULL OR TRIM(file_path) = ''")
           .where(acquisition_reservation_token: nil)
-          .update_all(file_path: imported_path, updated_at: Time.current)
+          .update_all(
+            file_path: imported_path,
+            reference_target_roots: reference_target_roots,
+            updated_at: Time.current
+          )
         unless claimed == 1
           raise BookAcquisitionConflictError,
             "Another acquisition claimed this title while post-processing was finalizing"
         end
         book.file_path = imported_path
+        book[:reference_target_roots] = reference_target_roots
       elsif book.file_path != imported_path
         raise BookAcquisitionConflictError,
           "Another acquisition already attached a different library file to this title"
@@ -400,7 +413,14 @@ class PostProcessingJob < ApplicationJob
     end
 
     snapshot = if source_stat.directory?
-      FileCopyService.snapshot_source_root(source_path)
+      if reference_completed_downloads?
+        FileCopyService.snapshot_reference_source_root(
+          source_path,
+          authorized_roots: shared_roots
+        )
+      else
+        FileCopyService.snapshot_source_root(source_path)
+      end
     elsif source_stat.file?
       FileCopyService.snapshot_source_file(source_path)
     elsif reference_link
@@ -424,7 +444,24 @@ class PostProcessingJob < ApplicationJob
       raise "Refusing to import source outside a configured download root."
     end
 
-    { directory: source_stat.directory?, snapshot: snapshot }
+    {
+      directory: source_stat.directory?,
+      snapshot: snapshot,
+      reference_regular_source_root: reference_completed_downloads? ?
+        reference_regular_source_root_for(snapshot, snapshotted_path, roots) : nil
+    }
+  end
+
+  def reference_regular_source_root_for(snapshot, source_path, source_roots)
+    return if snapshot.is_a?(FileCopyService::ReferenceSourceSnapshot)
+    if snapshot.is_a?(FileCopyService::SourceRoot) &&
+        snapshot.entries.values.none? { |manifest| manifest[2] == :file }
+      return
+    end
+
+    source_roots.select do |root|
+      path_inside_root?(source_path, root)
+    end.max_by(&:length)
   end
 
   def shared_download_roots(download)
@@ -566,6 +603,8 @@ class PostProcessingJob < ApplicationJob
     source_authorization ||= authorize_import_source(source)
     directory_source = source_authorization.fetch(:directory)
     source_snapshot = source_authorization.fetch(:snapshot)
+    @completed_reference_target_roots = []
+    @reference_regular_source_root = source_authorization[:reference_regular_source_root]
     @import_source_root = source_snapshot if directory_source
     @import_source_file_snapshot = source_snapshot unless directory_source
     @imported_renamed_files = []
@@ -655,6 +694,7 @@ class PostProcessingJob < ApplicationJob
     remove_instance_variable(:@import_base_path) if instance_variable_defined?(:@import_base_path)
     remove_instance_variable(:@import_source_root) if instance_variable_defined?(:@import_source_root)
     remove_instance_variable(:@import_source_file_snapshot) if instance_variable_defined?(:@import_source_file_snapshot)
+    remove_instance_variable(:@reference_regular_source_root) if instance_variable_defined?(:@reference_regular_source_root)
     remove_instance_variable(:@verified_library_snapshots) if instance_variable_defined?(:@verified_library_snapshots)
     remove_instance_variable(:@require_durable_import) if instance_variable_defined?(:@require_durable_import)
   end
@@ -791,7 +831,7 @@ class PostProcessingJob < ApplicationJob
     paths = ebook_directory_files(source)
 
     paths.each do |path|
-      if File.symlink?(path)
+      if File.symlink?(path) && !reference_source_snapshot(path)
         raise "Unsupported ebook import file type: #{File.basename(path)}"
       end
 
@@ -816,6 +856,14 @@ class PostProcessingJob < ApplicationJob
   end
 
   def allowed_ebook_import_file?(path, extension: nil)
+    if (snapshot = reference_source_snapshot(path)).is_a?(FileCopyService::ReferenceSourceSnapshot)
+      target = snapshot.target_snapshot.path.to_s
+      extension ||= File.extname(path).delete_prefix(".").downcase
+      return false unless File.file?(target) && !File.symlink?(target)
+
+      return EBOOK_ALLOWED_EXTENSIONS.include?(extension) &&
+        valid_ebook_import_content?(target, extension)
+    end
     return false if File.symlink?(path)
     return false unless File.file?(path)
 
@@ -895,12 +943,13 @@ class PostProcessingJob < ApplicationJob
 
   def import_file(source, destination)
     if reference_completed_downloads?
+      source_snapshot = reference_source_snapshot(source)
       FileCopyService.reference_noreplace(
         source,
         destination,
         root: @import_base_path,
-        source_root: @import_source_root,
-        source_snapshot: @import_source_file_snapshot
+        source_root: source_snapshot ? nil : @import_source_root,
+        source_snapshot: source_snapshot
       )
       @referenced_file_count += 1
     elsif hardlink_completed_downloads?
@@ -1025,12 +1074,13 @@ class PostProcessingJob < ApplicationJob
   end
 
   def reference_target_matches?(source, destination)
+    source_snapshot = reference_source_snapshot(source)
     FileCopyService.reference_target_matches?(
       source,
       destination,
       root: @import_base_path,
-      source_root: @import_source_root,
-      source_snapshot: @import_source_file_snapshot
+      source_root: source_snapshot ? nil : @import_source_root,
+      source_snapshot: source_snapshot
     )
   end
 
@@ -1059,7 +1109,11 @@ class PostProcessingJob < ApplicationJob
   def import_directory_entry(source, destination)
     stat = File.lstat(source)
     if stat.symlink?
-      raise "Refusing to import symbolic link: #{source}"
+      unless reference_completed_downloads? && reference_source_snapshot(source)
+        raise "Refusing to import symbolic link: #{source}"
+      end
+
+      import_file_without_duplicate_content(source, File.join(destination, File.basename(source)))
     elsif stat.directory?
       nested_destination = File.join(destination, File.basename(source))
       ensure_real_import_directory!(nested_destination)
@@ -1180,17 +1234,42 @@ class PostProcessingJob < ApplicationJob
   end
 
   def record_imported_source!(source)
+    record_reference_target_root!(source) if reference_completed_downloads?
     return unless @import_source_root
 
     relative = Pathname(source).expand_path.relative_path_from(@import_source_root.path)
     manifest = @import_source_root.entries[relative.to_s]
-    unless manifest && manifest[2] == :file
+    unless manifest && manifest[2].in?([ :file, :reference ])
       raise FileCopyService::UnsafePathError, "imported source is absent from the immutable manifest"
     end
 
     @imported_source_files << relative.to_s
   rescue ArgumentError
     raise FileCopyService::UnsafePathError, "imported source escaped the immutable manifest"
+  end
+
+  def record_reference_target_root!(source)
+    snapshot = reference_source_snapshot(source)
+    root = if snapshot.is_a?(FileCopyService::ReferenceSourceSnapshot)
+      snapshot.authorized_root.to_s
+    else
+      @reference_regular_source_root
+    end
+    @completed_reference_target_roots << root if root.present?
+    @completed_reference_target_roots.uniq!
+  end
+
+  def reference_source_snapshot(source)
+    expanded = Pathname(source).expand_path
+    if @import_source_file_snapshot && expanded == @import_source_file_snapshot.path
+      return @import_source_file_snapshot
+    end
+    return unless @import_source_root&.reference_snapshots
+
+    relative = expanded.relative_path_from(@import_source_root.path).to_s
+    @import_source_root.reference_snapshots[relative]
+  rescue ArgumentError
+    nil
   end
 
   def renamed_destination_file(source, destination, book)
--- a/app/jobs/post_processing_job.rb
+++ b/app/jobs/post_processing_job.rb
@@ -118,6 +118,7 @@ class PostProcessingJob < ApplicationJob
         require_durable: move_completed_downloads? || remove_usenet_download,
         source_authorization: source_authorization
       )
+      validate_completed_reference_target_roots! if reference_completed_downloads?
 
       book_path = imported_book_path(book, destination)
       cleanup_state = source_cleanup&.fetch(:state)
@@ -431,7 +432,7 @@ class PostProcessingJob < ApplicationJob
     snapshotted_path = if source_stat.directory?
       snapshot.canonical_path.to_s
     elsif reference_link
-      snapshot.target_snapshot.path.to_s
+      snapshot.canonical_target_path.to_s
     else
       snapshot.canonical_parent_path.join(snapshot.path.basename).to_s
     end
@@ -459,9 +460,10 @@ class PostProcessingJob < ApplicationJob
       return
     end
 
-    source_roots.select do |root|
+    root = source_roots.select do |root|
       path_inside_root?(source_path, root)
     end.max_by(&:length)
+    FileCopyService.snapshot_reference_root(root) if root
   end
 
   def shared_download_roots(download)
@@ -944,12 +946,15 @@ class PostProcessingJob < ApplicationJob
   def import_file(source, destination)
     if reference_completed_downloads?
       source_snapshot = reference_source_snapshot(source)
+      regular_root_snapshot = @reference_regular_source_root unless
+        source_snapshot.is_a?(FileCopyService::ReferenceSourceSnapshot)
       FileCopyService.reference_noreplace(
         source,
         destination,
         root: @import_base_path,
         source_root: source_snapshot ? nil : @import_source_root,
-        source_snapshot: source_snapshot
+        source_snapshot: source_snapshot,
+        authorized_root_snapshot: regular_root_snapshot
       )
       @referenced_file_count += 1
     elsif hardlink_completed_downloads?
@@ -1075,12 +1080,15 @@ class PostProcessingJob < ApplicationJob
 
   def reference_target_matches?(source, destination)
     source_snapshot = reference_source_snapshot(source)
+    regular_root_snapshot = @reference_regular_source_root unless
+      source_snapshot.is_a?(FileCopyService::ReferenceSourceSnapshot)
     FileCopyService.reference_target_matches?(
       source,
       destination,
       root: @import_base_path,
       source_root: source_snapshot ? nil : @import_source_root,
-      source_snapshot: source_snapshot
+      source_snapshot: source_snapshot,
+      authorized_root_snapshot: regular_root_snapshot
     )
   end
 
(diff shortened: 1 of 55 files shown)


Pedro-Revez-Silva (commented):
No verified findings.

Reviewed exact head `ee5b3a38675949e883b445fb648fc8fcb6c7851e` against merge-base `ba97a841855026149fe4e3bb2f7ad0ead922da5e` through independent acquisition, reference-security, filesystem-recovery, test, migration, and dependency passes.

Validation completed:
- full Rails suite: 3,130 tests, 12,023 assertions
- system suite: 33 tests, 260 assertions
- line coverage 91.35%, branch coverage 73.40%
- RuboCop, Brakeman, bundler-audit, CodeQL, companion smoke, amd64/ARM filesystem contracts, and production container build all pass
- no existing review threads or duplicate findings

Residual platform risk is limited to real-world DrvFS and NFS behavior that cannot be reproduced faithfully on this runner; the PR exercises those branches with deterministic mount metadata, real local syscalls, race injection, and CIFS contracts on both architectures.

Review state is COMMENT because GitHub does not permit approving one’s own pull request.

app/services/release_scorer.rb
```diff
@@ -208,6 +231,160 @@ def usenet?
...
+    )
+    return unless series_match
+
+    series_range = series_match.begin(:series)...series_match.end(:series)
+    tail_offset = series_match.end(:series)
+    tail = title[tail_offset..]
+    marker = '(?:#|(?<![[:alnum:]])(?:issues?|no\.?|numbers?)(?![[:alnum:]])\s*#?\s*)'
+    issue_value = COMIC_ISSUE_VALUE_PATTERN
+    issue = "(?<issue>#{issue_value})"
+    issue_terminator = '(?=\z|\s|[\(\[])'
```
Pedro-Revez-Silva: [P2] Accept normal release delimiters after comic issue tokens

The terminator only permits end-of-string, whitespace, `(`, or `[`, but torrent/indexer titles routinely separate fields with `.` or `_`. On this exact head, `Saga #054 English Comic CBZ` is an exact match scoring 100, while `Saga.#054.English.Comic.CBZ` and `Saga_#054_English_Comic_CBZ` are classified as unknown, score below 50, and cannot auto-select. Because unknown comic matches are capped at 49, otherwise exact releases miss the normal confidence threshold. Please accept standard release delimiters after a complete issue token while continuing to reject suffix/range/list continuations.

app/services/release_scorer.rb
```diff
@@ -208,6 +231,160 @@ def usenet?
...
+    end
+  end
+
+  def additional_comic_issue_value?(value, run_years)
+    !value.match?(/\A(?:19|20)\d{2}\z/) || !run_years.include?(value.to_i)
+  end
+
+  def comic_issue_run_year_conflict?(detected)
+    requested_year = @book.release_date&.year
+    detected[:run_year].present? && requested_year.present? && detected[:run_year] != requested_year
```
Pedro-Revez-Silva: [P2] Compare a release's run year with volume metadata, not the issue cover date

`ComicVineClient` populates an issue book's `release_date` from `cover_date`, while the year encoded in comic release names is normally the volume/run start year. For example, Saga volume 1 began in 2012 but issue 54 has a 2018 cover date; `Saga (2012) #54 English Comic CBZ` is therefore downgraded to unknown and capped below 50 on this head. Later issues from long-running or relaunched series will be systematically rejected. Please persist/use the volume `start_year`, or avoid treating a lone run-year difference from an issue cover date as a conflict.

app/services/direct_download_file_service.rb
```diff
@@ -91,6 +100,23 @@ def cleanup_orphans!(root:, max_age: ORPHAN_MAX_AGE.ago)
       0
     end
 
+    def legacy_staging_diagnostic(root:)
+      root = Pathname(root).expand_path
+      legacy = root.join(LEGACY_STAGING_DIRECTORY)
+      stat = File.lstat(legacy)
+      mode = stat.mode & 0o7777
+      return if stat.directory? && mode == 0o700
```
Pedro-Revez-Silva: [P2] Surface retained legacy payloads before dropping their only owner

A normal pre-upgrade `.shelfarr-staging` directory is mode 0700, so this returns healthy even when it still contains an interrupted direct-download tree. In a full reconciliation repro, `remove_staging!` retained those bytes but returned success, after which the reservation and all recovery columns were cleared; the bytes remained and `legacy_staging_diagnostic` returned nil. The v2 orphan sweep never scans the legacy namespace, leaving a silent, potentially large disk leak with no DB owner or health signal. Please diagnose non-empty legacy direct-download payloads (or persist an explicit manual-review marker) before clearing their recovery state.

app/jobs/post_processing_job.rb
```diff
@@ -96,11 +97,15 @@ def perform_locked(download, source_path_retry_count, expected_owner_job_id)
...
-        source_resolution[:authorized_roots],
-        download
-      )
+      source_authorization = begin
+        validate_download_specific_source_path!(
+          source_path,
+          source_resolution[:authorized_roots],
+          download
+        )
+      rescue FileCopyService::ReferenceTargetUnavailableError
```
Pedro-Revez-Silva, line 106: [P2] Retry reference targets that disappear after authorization

This rescue only covers the initial authorization snapshot. If a debrid/rclone symlink target disappears after this succeeds but before `import_files` opens it, the `File.exist?` check there raises the generic source-not-found error instead. A focused full-job harness with `post_processing_source_path_retries = 2` removed the target immediately after authorization and verified that no retry was enqueued and the request was marked attention-needed. Please preserve the typed transient-target handling through the open/import boundary, at least until publication begins.

Pedro-Revez-Silva (commented):
Reviewed exact head `ee5b3a38675949e883b445fb648fc8fcb6c7851e` against base `ba97a841855026149fe4e3bb2f7ad0ead922da5e`. The findings below are blocking: GitHub does not allow an author to request changes on their own PR, so this is submitted as a comment, but my merge recommendation is **not ready** until they are addressed.

Validation: all remote checks are green on this head. The local system suite passed (33 runs / 260 assertions), and manual browser QA covered sign-in, dashboard, request listing/filtering, comic detail/search-result selection, completed-request detail, and the download action with no console errors. The macOS local gate reached Rails tests and reported two new errors from the direct-download test file; its one additional failure reproduces on the base commit and is not attributed to this PR.

app/services/file_copy_service.rb
```diff
@@ -2337,6 +2904,87 @@ def persist_copy_lock_identity!(lock, token, identity)
...
+      journal = File.for_fd(descriptor, "r+b", autoclose: true)
+      begin
+        raise UnsafePathError, "DrvFS copy journal is not a regular file" unless journal.stat.file?
+
+        acquired = begin
+          journal.flock(File::LOCK_EX | File::LOCK_NB)
+        rescue Errno::EWOULDBLOCK
+          false
+        end
+        until acquired
```
Pedro-Revez-Silva, line 2947: [P2] Bound waits on the root-wide DrvFS copy journal

This root-wide lock loop has no deadline or cancellation path. A live worker stalled in destination I/O can therefore hold the journal indefinitely while every later import waits here; `PostProcessingJob` calls `cp_noreplace` without a heartbeat, and the default queue has only three worker threads, so one holder plus two waiters can exhaust all workers. A lock-contention harness remained in this loop with repeated heartbeats until externally stopped. Please add a bounded/cancellable wait and raise a recoverable busy error so the job can be retried instead of occupying workers forever.

test/services/direct_download_file_service_test.rb
```diff
@@ -55,6 +55,156 @@ class DirectDownloadFileServiceTest < ActiveSupport::TestCase
...
+    legacy = File.join(@output_root, DirectDownloadFileService::LEGACY_STAGING_DIRECTORY)
+    legacy_marker = File.join(legacy, "retained-legacy-entry")
+    FileUtils.mkdir_p(legacy)
+    File.binwrite(legacy_marker, "legacy bytes")
+    File.chmod(0o777, legacy)
+    legacy_stat = File.stat(legacy)
+    chmod_paths = []
+    real_fchmod = FileCopyService.method(:native_fchmod)
+    guarded_fchmod = lambda do |descriptor, mode|
+      path = File.readlink("/proc/self/fd/#{descriptor}").delete_suffix(" (deleted)")
```
Pedro-Revez-Silva: [P2] Keep the declared local gate portable on macOS

This new test unconditionally reads Linux-only `/proc/self/fd`, producing `ENOENT` on macOS. The same file has a second deterministic macOS error because `with_forced_drvfs_mount` only stubs mountinfo while production `drvfs_mount?` returns false before reading it on non-Linux platforms, so the symlink case takes generic cleanup and raises `ELOOP`. Both errors occur in the focused file and the full `bin/quality push` Rails run. Please gate Linux-specific cases or use platform-neutral descriptor/mount seams (including stubbing the platform predicate) so the documented local gate remains runnable on supported developer machines.

app/services/release_scorer.rb
Pedro-Revez-Silva: Addressed in 4a30c96: standard dot and underscore delimiters now terminate complete comic issue tokens while range and list continuations remain rejected.

app/services/release_scorer.rb
Pedro-Revez-Silva: Addressed in 4a30c96 and hardened in 9fcc373: scoring uses persisted volume start-year metadata; existing Comic Vine issue rows are selected for backfill, and explicit run years fail safe until that metadata is available.

app/services/direct_download_file_service.rb
Pedro-Revez-Silva: Addressed in e55c18b: retained legacy staging payloads preserve their recovery owner and surface a health diagnostic instead of being silently orphaned.

app/jobs/post_processing_job.rb
Pedro-Revez-Silva: Addressed in d64d512: target loss after authorization remains a typed retryable reference-target failure through the import boundary.

app/services/file_copy_service.rb
Pedro-Revez-Silva: Addressed in d169b68: root-journal waits are heartbeat-aware and bounded, returning a recoverable busy error instead of occupying workers indefinitely.

test/services/direct_download_file_service_test.rb
Pedro-Revez-Silva: Addressed in e55c18b: descriptor validation no longer hardcodes /proc/self/fd, and the DrvFS behavior is exercised through an explicit platform seam; the macOS local gate now passes.

Pedro-Revez-Silva (commented):
Exact-head review complete.

Base: ba97a841855026149fe4e3bb2f7ad0ead922da5e
Head: 9fcc3733701ec573f112bd7efe8a031b9026d01d

No remaining verified code findings. The two upgrade-path issues found during the final pass were fixed in 9fcc373: existing Comic Vine issue rows now receive run-year backfill coverage and explicit run-year releases fail safe until that metadata exists; the Libation container smoke script now supports both GNU and BSD date.

Validation on this exact head:
- focused regression tests: 38 tests, 230 assertions
- local bin/quality push: passed
- coverage: 91.19% line, 73.28% branch
- manual local UI flow: passed with a clean browser console
- GitHub CI, filesystem contracts, companion/container jobs, security, and CodeQL: all passed
- all review conversations: resolved
- commit signature: verified

GitHub does not permit the PR author to approve their own PR, so the remaining review requirement needs another eligible reviewer or an authorized admin merge.

Pedro-Revez-Silva (commented):
Exact-head follow-up review complete.

Base: ba97a841855026149fe4e3bb2f7ad0ead922da5e
Head: e70eb5c491cbdb841f9f72e004cf98aef76c5513

No remaining verified code or UI findings. Comic books now render as Comics & Manga with the established emerald treatment on dashboard recent-library and recent-request cards, the requests list, library cards, and library details.

Validation on this exact head:
- focused rendering suite: 187 tests, 883 assertions
- local bin/quality push: passed twice, including the enforced pre-push gate
- coverage: 91.19% line, 73.28% branch
- manual connected-browser UI verification: passed on all affected surfaces
- browser console and page logs: clean
- GitHub CI, filesystem contracts, companion/container jobs, security, and CodeQL: all passed
- all review conversations: resolved
- commit signature: verified

GitHub does not permit the PR author to approve their own PR, so the remaining review requirement still needs another eligible reviewer or an authorized admin merge.

PR #508, opened 2026-09-01
Handle combined-title search safely across providers

### LINKED ISSUE #507: Combined localized/original titles produce incorrect indexer searches
## Description

Shelfarr appears to use a combined localized/original title as the search title when querying indexers.

For example, the book **La Nena** by Carmen Mola was stored in the database as:

```
La Nena / The Girl
```

The record is:

```ruby
Book.where("title LIKE ?", "%La Nena%").pluck(:id, :title, :author, :metadata_source)
# => [[8, "La Nena / The Girl", "Carmen Mola", "openlibrary"]]
```

I tested this by temporarily changing only the `Book.title` value from:

```
La Nena / The Girl
```

to:

```
La Nena
```

After doing so, the exact same search immediately returned results from Jackett:

```
08-31 21:50:00 Info Torznab search in tag:books for La Nena ES => Found 8 releases [3ms]
```

This strongly suggests that the problem is the search query generated from the stored `Book.title`, rather than Jackett being unable to find the book.

Importantly, this does **not** establish where the combined title `La Nena / The Girl` originates. The metadata source is Open Library, but it is not clear from the current investigation whether Open Library returns that exact combined value or whether Shelfarr constructs it while processing metadata.

## Current behavior

The relevant search code in `app/jobs/search_job.rb` uses `book.title` directly:

```ruby
query_parts = [ book.title ]
query_parts << book.author if book.author.present?
query = query_parts.join(" ")
```

Therefore, when the stored title is:

```
La Nena / The Girl
```

Shelfarr uses that combined value as the basis for the indexer search.

The corresponding Jackett log for the problematic search was:

```
08-31 20:40:15 Info Torznab search in tag:books for La Nena / The Girl ES => Found 0 releases [10678ms]
```

For comparison, after changing the title to `La Nena`, Jackett logged:

```
08-31 21:50:00 Info Torznab search in tag:books for La Nena ES => Found 8 releases [3ms]
```

## Expected behavior

Localized and original/international titles should not necessarily be treated as one literal search title.

If metadata contains both:

```
La Nena / The Girl
```

Shelfarr should ideally handle these as separate title variants when constructing indexer searches, or otherwise determine the appropriate title for the search based on the book's language/context.

For example, a search for the Spanish edition should be able to use:

```
La Nena
```

rather than requiring an indexer to match the combined string:

```
La Nena / The Girl
```

The external metadata should not need to be modified. The issue appears to be how Shelfarr uses the title when constructing the search query.

## Reproduction

1. Import or obtain the metadata for **La Nena** by Carmen Mola.
2. Observe that `Book.title` is:

   ```
   La Nena / The Girl
   ```

3. Run the indexer search using Jackett.
4. Observe that the search does not return the expected releases.
5. Temporarily change only the database value:

   ```ruby
   book = Book.find(8)
   book.update!(title: "La Nena")
   ```

6. Run the exact same search again.
7. Jackett now returns results:

   ```
   Torznab search in tag:books for La Nena ES => Found 8 releases
   ```

## Additional observation

A different book by the same author, **La novia gitana**, is stored normally:

```ruby
Book.where("title LIKE ?", "%La novia gitana%").pluck(:id, :title, :author, :metadata_source)
# => [[6, "La novia gitana", "Carmen Mola", "openlibrary"]]
```

This suggests the problem is not simply related to the author or the general Jackett search configuration.

## Question

Could the title handling and indexer query construction be reviewed so that localized/original title combinations such as `La Nena / The Girl` do not get sent as a single literal search title?

Ideally, Shelfarr could preserve the metadata as provided while using an appropriate title variant, or multiple title variants, when searching indexers.

Pedro-Revez-Silva:
## Assigned issue

Closes #507

## What changed

Searches for metadata titles such as `La Nena / The Girl` now treat the localized component, original component, and combined text as bounded fallback variants without changing the stored book title.

- Indexer attempts prioritize exact and title-plus-author queries for every variant, then broaden only the preferred title.
- Primary attempts are capped at 18 and category-free broad attempts at 6 so aliases and numeric rewrites cannot exceed the search lease.
- Broad searches interleave language-tagged and hint-free variants and preserve compact comic issue queries.
- Anna's Archive, Z-Library, LibriVox, Project Gutenberg, and store providers try the next variant only when the earlier one returns no results.
- Release scoring uses phrase boundaries, so short titles such as `It` do not match `The Institute`.
- Because a spaced slash can also describe an omnibus, component-only results remain available for manual review but cannot be auto-selected; a release containing the full combined title can still auto-select.

## Verification

- Adversarial autoselection checks for short aliases and ambiguous `Hamlet / Macbeth` metadata.
- Non-English category-free fallback proved an untagged original-title result is reached under the cap.
- Comic primary and broad searches both retain `Saga 7`, `Saga #7`, and `Saga 007` variants.
- Single-query providers stop after the first non-empty variant.
- Focused suite: 135 tests, 592 assertions, all passing.
- Full local quality gate passed, including system tests.
- RuboCop and diff checks passed.

## Screenshots or recordings

Not applicable; this changes backend query planning and scoring without changing rendered UI.

## Checklist

- [x] The maintainer assigned the linked issue before implementation
- [x] The change stays within the issue's scope
- [x] Regression tests cover the changed behavior
- [x] Relevant local quality checks passed
- [x] No user-facing documentation change is required

### DIFF
--- a/app/jobs/search_job.rb
+++ b/app/jobs/search_job.rb
@@ -18,6 +18,8 @@ class SearchJob < ApplicationJob
   # otherwise come back empty, so users see low-confidence candidates
   # instead of no results at all.
   LOW_CONFIDENCE_FALLBACK_LIMIT = 5
+  MAX_GENERIC_SEARCH_ATTEMPTS = 18
+  MAX_BROAD_SEARCH_ATTEMPTS = 6
 
   ROMAN_NUMERALS = {
     "I" => "1",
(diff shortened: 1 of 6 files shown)


Pedro-Revez-Silva (commented):
Adversarial review complete on exact head 277203b8615480413575ea84261696977751a95f against base 2c91db2238f638ed2f0fd59673e376ec7695c7f7. No findings remain. Verified bounded provider fallback, combined-title component releases remain manual-only, full-title matches can auto-select, non-English and comic variant coverage, focused/full suites, the combined integration gate, container/filesystem checks, and all CI checks.

PR #516, opened 2026-09-02
Optimize AudiobookshelfLibraryMatcherService to eliminate O(N*M) text processing overhead

### LINKED ISSUE #511: Library fuzzy matcher recomputes normalization and trigrams per compared pair — admin requests index and search enrichment can exceed proxy timeouts on large libraries
## Environment

- shelfarr v2026.08.31.1 (docker, `ghcr.io/pedro-revez-silva/shelfarr:latest`, digest `08e06f5b…`)
- Synology DS1815+ (Intel Atom C2538 @ 2.4 GHz), SQLite, nginx in front with a 60 s upstream timeout
- 9,115 `library_items` synced from Audiobookshelf (8,359 + 756 across two libraries), 6 requests

## Symptom

- The **requests index as an admin** takes 20 s+ and intermittently exceeds the 60 s proxy timeout (502/504). Non-admin index is instant.
- The **search results page is slow for every user** once results land.
- Login and other pages respond in 30–55 ms, so this is page-specific, not host load.

## Root cause

`AudiobookshelfLibraryMatcherService#matches_for` redoes all string preparation **per compared pair**: for every library item it calls `normalize_text` (UTF-8 re-encode, NFKD, `\p{Mn}` strip, case fold, two more regex passes) up to three times and rebuilds trigram `Set`s for both sides, every time. Nothing is reused across items, queries, or calls.

The call sites multiply that cost:

- `RequestsController#index` (admin only): `matches_for` per request book → requests × items pairs (6 × 9,115 ≈ 55k here).
- `SearchController` enrichment: `matches_for_many(results, limit_per_result: 3)` → results × items pairs (20 × 9,115 ≈ 182k).
- `RequestsController#new` / `create`: `load_synced_library_matches` scans the full inventory per scope with `cache_library_items: false`.

## Measurements

I extracted the hot path (`normalize_text` → `trigrams` → `trigram_similarity` → `match_score`, verbatim) into a standalone script and ran it against the real 9,108-row inventory:

| workload | pairs | Apple M-series | est. Atom C2538 (~10–20× slower single-thread) |
|---|---|---|---|
| admin requests index (5 books) | 45,540 | 1.13 s | ~11–23 s |
| search page (20 results) | 182,160 | 4.44 s | ~45–90 s |

That matches the observed behavior: the admin index sits around the timeout boundary and tips over whenever anything else (e.g. the hourly library sync) contends.

## Proposed fix (patch ready)

Behavior-preserving refactor of the service only:

1. Precompute each item's normalized titles/author and their trigram sets **once**, memoized on the matcher instance keyed by `[id, updated_at]` (so streamed re-reads that changed are recomputed; unsaved items skip the memo).
2. Compute each query's trigram sets once per `matches_for` call and pass sets into `trigram_similarity` instead of raw strings.
3. In `trigram_similarity`, count the intersection by probing the smaller set and derive the union by inclusion–exclusion instead of allocating `&` and `|` sets.

Scoring is bit-identical: replayed all 182,160 real pairs through both implementations — **0 score mismatches, 7.1× faster** on the search-page workload (the win compounds because `matches_for_many` reuses one matcher across results). Larger algorithmic wins (an inverted trigram index for candidate pruning) are possible but out of scope for a minimal reviewable change.

I have the patch ready on a fork, including two added tests (memo reuse across queries on one instance; streaming matcher seeing item updates between queries), with `bin/quality push` run against it. Happy to open the PR if you assign this issue per CONTRIBUTING.md.

Pedro-Revez-Silva:
## Assigned issue

Closes #511

## What changed

Behavior-preserving performance refactor of `AudiobookshelfLibraryMatcherService` only. Controllers, routes, and scoring weights are unchanged.

On large synced libraries, admin request index and search enrichment called `matches_for` / `matches_for_many` against the full inventory and re-normalized titles/authors and rebuilt trigram sets for every pair. This PR:

- Precomputes one prepared query (normalized title/author + trigram sets)
- Precomputes one prepared item per library row, including the combined title+subtitle display string and its trigram set, keyed by `[id, title, subtitle, author]`
- Reuses that prep across queries on the same matcher instance (`matches_for_many`, admin request index)
- Passes those sets into the original `match_score` / `trigram_similarity` path, including the string-equality short-circuit
- Counts trigram overlap instead of allocating full intersection/union sets

A first patch used a parallel precomputed scorer that unique'd trigram sets rather than title strings, and it still rebuilt display-title trigrams on every query. This revision removes that second path so scoring cannot drift, and caches the combined title/subtitle representation once per item.

## Verification

**Tests added/extended:**
- Same matcher instance reuses item prep across queries (counts `normalize_text` / `trigrams`; second query stays well below a full rescan)
- One matcher is cheaper than constructing a matcher per query (the `matches_for_many` pattern)
- Cached-mode matcher refreshes prep when the in-memory item it iterates is updated
- Streaming matcher (`cache_library_items: false`) refreshes after a persisted update
- Optimized match lists stay equivalent to a copy of the original algorithm for the same inputs
- Precomputed trigram overlap matches the original `&` / `|` set-algebra scores

**Tests run:**
```
PARALLEL_WORKERS=1 bin/rails test test/services/audiobookshelf_library_matcher_service_test.rb
# 25 runs, 85 assertions, 0 failures, 0 errors

PARALLEL_WORKERS=1 bin/rails test \
  test/services/audiobookshelf_library_matcher_service_test.rb \
  test/controllers/requests_controller_test.rb:2904 \
  test/controllers/search_controller_rendering_test.rb:134
# 27 runs, 93 assertions, 0 failures, 0 errors

bin/rubocop --force-exclusion \
  app/services/audiobookshelf_library_matcher_service.rb \
  test/services/audiobookshelf_library_matcher_service_test.rb
```

Exact Synology / Atom / reverse-proxy wall-clock 502/504 timing was not reproduced 1:1 here. The algorithmic waste is covered by the reuse tests; scoring is locked against the original algorithm.

## Checklist

- [x] The maintainer assigned the linked issue to me before I started implementation
- [x] This pull request stays within the assigned issue's scope
- [x] I added or updated tests for the behavior I changed
- [x] I ran the relevant local quality checks
- [x] I updated documentation for user-visible changes (N/A — internal matcher optimization)

<div><a href="https://cursor.com/agents/bc-46d51315-7ea9-4f8f-b895-b7108c139e3a?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-46d51315-7ea9-4f8f-b895-b7108c139e3a&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/app/services/audiobookshelf_library_matcher_service.rb
+++ b/app/services/audiobookshelf_library_matcher_service.rb
@@ -17,10 +17,18 @@ class AudiobookshelfLibraryMatcherService
 
   FuzzyThreshold = 85
   MaxMatchTextLength = 500
+  TrigramCodepointBase = 0x110000
+  EmptyTrigrams = [].freeze
+  # A matcher can scan thousands of externally sourced records. Bound retained
+  # trigram entries so pathological metadata cannot turn this speed cache into
+  # a per-request memory spike; uncached entries still use identical scoring.
+  MaxCachedTrigrams = 1_000_000
+  private_constant :TrigramCodepointBase, :EmptyTrigrams, :MaxCachedTrigrams
 
   def initialize(cache_library_items: true)
     @cache_library_items = cache_library_items
     @prepared_items = {}
+    @cached_item_trigram_count = 0
   end
 
   def self.matches_for_many(results, limit_per_result: 3)
@@ -121,6 +129,8 @@ class AudiobookshelfLibraryMatcherService
   end
 
   def prepared_item(item)
+    return build_prepared_item(item, retain_trigrams: false) unless @cache_library_items
+
     key = item_prep_cache_key(item)
     return @prepared_items[key] if @prepared_items.key?(key)
 
@@ -131,7 +141,7 @@ class AudiobookshelfLibraryMatcherService
     [ item.id, item.title, item.subtitle, item.author ]
   end
 
-  def build_prepared_item(item)
+  def build_prepared_item(item, retain_trigrams: true)
     return nil if oversized_match_text?(item.title)
 
     item_title = normalize_text(item.title)
@@ -141,13 +151,43 @@ class AudiobookshelfLibraryMatcherService
     return nil if item_title.blank? && item_display_title.blank? && item_author.blank?
 
     titles = [ item_title, item_display_title ].uniq
+    title_trigrams_by_title, author_trigrams = if retain_trigrams
+      cacheable_item_trigrams(titles, item_author)
+    else
+      item_trigrams(titles, item_author)
+    end
     PreparedItem.new(
       title: item_title,
       titles: titles,
       author: item_author,
-      title_trigrams_by_title: titles.to_h { |title| [ title, title.present? ? trigrams(title) : Set.new ] },
-      author_trigrams: item_author.present? ? trigrams(item_author) : nil
+      title_trigrams_by_title: title_trigrams_by_title,
+      author_trigrams: author_trigrams
     )
+  end
+
+  def cacheable_item_trigrams(titles, author)
+    estimated_count = titles.sum { |title| estimated_trigram_count(title) } + estimated_trigram_count(author)
+    return [ nil, nil ] if @cached_item_trigram_count + estimated_count > max_cached_trigrams
+
+    title_trigrams, author_trigrams = item_trigrams(titles, author)
+    @cached_item_trigram_count += title_trigrams.sum { |_title, values| values.size }
+    @cached_item_trigram_count += author_trigrams.size if author_trigrams
+    [ title_trigrams, author_trigrams ]
+  end
+
+  def item_trigrams(titles, author)
+    [
+      titles.to_h { |title| [ title, title.present? ? trigrams(title) : EmptyTrigrams ] },
+      author.present? ? trigrams(author) : nil
+    ]
+  end
+
+  def estimated_trigram_count(text)
+    text.present? ? text.length + 2 : 0
+  end
+
+  def max_cached_trigrams
+    MaxCachedTrigrams
   end
 
   def match_score(query_title:, query_author:, item_titles:, item_author:,
@@ -204,14 +244,36 @@ class AudiobookshelfLibraryMatcherService
   end
 
   def trigram_overlap_size(left, right)
-    smaller, larger = left.size <= right.size ? [ left, right ] : [ right, left ]
-    smaller.count { |trigram| larger.include?(trigram) }
+    left_index = 0
+    right_index = 0
+    overlap = 0
+
+    while left_index < left.size && right_index < right.size
+      comparison = left[left_index] <=> right[right_index]
+      if comparison.negative?
+        left_index += 1
+      elsif comparison.positive?
+        right_index += 1
+      else
+        overlap += 1
+        left_index += 1
+        right_index += 1
+      end
+    end
+
+    overlap
   end
 
   def trigrams(text)
-    return Set.new if text.blank?
+    return EmptyTrigrams if text.blank?
 
-    padded = "  #{text}  "
-    (0..padded.length - 3).map { |i| padded[i, 3] }.to_set
+    codepoints = "  #{text}  ".codepoints
+    values = Array.new(codepoints.length - 2) do |index|
+      ((codepoints[index] * TrigramCodepointBase) + codepoints[index + 1]) * TrigramCodepointBase +
+        codepoints[index + 2]
+    end
+    values.sort!
+    values.uniq!
+    values.freeze
   end
 end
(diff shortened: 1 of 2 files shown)


severoon:
Validated this branch against the deployment that motivated #511 (9,108 real `library_items` from a Synology DS1815+ / Atom C2538 install).

**Method:** ran this branch's `AudiobookshelfLibraryMatcherService` and current `main`'s over the same 9,108 items with 20 search-shaped queries (the `matches_for_many` pattern), recording every `match_score` result plus the returned match lists.

**Results:**
- 182,160 scored pairs: **0 score mismatches** vs `main`
- returned match lists (items, scores, likely/possible labels): **identical**
- wall clock for the 20-query workload: 6.34 s → 1.01 s (**6.3×**) on an M-series Mac; the Atom that was timing out is ~10–20× slower single-thread, so this should bring the admin requests index and search enrichment well under the 60 s proxy timeout there
- this branch's matcher test file: 25 runs, 0 failures locally (Ruby 3.3.6)

**One observation from profiling on the same data:** after this change, ~90 % of a cold call is `build_prepared_item` over the full inventory (0.27 s of a 0.30 s single-query call on the Mac), and that cost recurs on every render because the memo lives on the per-request matcher instance — on the Atom-class NAS that's still roughly 5–9 s per admin requests page. Two possible follow-ups, in increasing order of impact/effort:

1. Promote the prepared-item cache to process level — the `[id, title, subtitle, author]` key already makes entries self-invalidating — leaving only the per-pair intersections (~2–4 s on the same hardware).
2. Since a passing score ≥ 85 requires best-title Jaccard ≥ (85 − 0.3·100)/0.7 ≈ 0.786, candidates can be pruned hard with an inverted trigram index before scoring — sub-second on the same hardware, still score-identical.

Happy to file either as a follow-up issue, and to re-test the final revision or released image on the same hardware if useful.

Pedro-Revez-Silva (commented):
Adversarial review complete on exact head  against base . No findings remain. Verified bounded-memory Unicode trigram matching, exact-score fallback, 10,000-case randomized equivalence, focused and full Rails quality suites, static analysis, and all required CI checks.

Pedro-Revez-Silva:
Exact reviewed head: 0632580db913a4da4309fe1c509f2e3e7abed1d7. Pinned base during review: 2c91db2238f638ed2f0fd59673e376ec7695c7f7. This supplies the SHAs omitted from the review text; the submitted review itself was attached to the correct commit ID.

severoon:
Following up on the offer above to re-test the released image on the hardware behind #511 (Synology DS1815+, Atom C2538, 9,115 synced `library_items`). Now running **2026.09.18.1**.

**Admin requests index, server-side, from the Rails log** (`RequestsController#index` as an admin, 51 queries per render):

| | Completed in |
|---|---|
| five unfiltered renders | 4,676 · 4,695 · 4,746 · 6,124 · 6,745 ms |
| one filtered render (34 queries) | 3,938 ms |
| same action as a non-admin, for scale | 37–143 ms |

Before the fix this page never finished: it ran past nginx's 60 s proxy timeout every time. So on this box the change is from "unusable" to roughly **5–7 s**, which is inside the 5–9 s I estimated above from the Mac profile. Thank you — the admin side of the app is usable again here.

Where the remaining time goes, from the same log lines: views ~60–130 ms, ActiveRecord ~0.4–0.55 s, GC ~0.45–0.8 s, and the other ~3.5–5 s is Ruby in the action. I have not profiled the released image, but that is consistent with the earlier observation that the per-item preparation is rebuilt on every render because the memo lives on the per-request matcher instance.

No action needed from this comment. For whenever large libraries come up again, the two ideas from my earlier comment still look like the places to go, and these numbers suggest what each would buy on hardware like this (estimates scaled from the Mac profile, not measured here):

1. A process-level cache for the prepared items (the `[id, title, subtitle, author]` key already makes entries self-invalidating): removes the per-render preparation and leaves the per-pair scoring, roughly 5–7 s → 2–4 s.
2. An inverted trigram index to prune candidates before scoring (a passing score needs a best-title Jaccard of about 0.79, so most of the inventory can be skipped, score-identical): roughly sub-second.

5–7 s on this Atom is a fine place to leave it, so take them or leave them as you see fit.

PR #518, opened 2026-09-04
Fix Solid Queue workers never starting with SOLID_QUEUE_IN_PUMA

### LINKED ISSUE #517: Background jobs (SearchJob) enqueue successfully but never execute — solid_queue_processes table always empty despite SOLID_QUEUE_IN_PUMA=1
**Environment:**
- Deployment: Docker on Unraid, manually configured container (fields set individually via Unraid's UI, equivalent to docker-compose.example.yml)
- Image: ghcr.io/pedro-revez-silva/shelfarr:latest
- Volumes: `/rails/storage` (persistent app storage) and a shared media/downloads mount
- Env vars set: `PUID=99`, `PGID=100`, `RAILS_ENV=production`, `SOLID_QUEUE_IN_PUMA=1`

**Summary:**
Requests are created successfully, the search UI works, and Prowlarr/Hardcover connections test successfully from the admin settings page. But the actual background job responsible for finding and grabbing a release (SearchJob) never executes. It shows as enqueued in the log but no "Performing"/"Performed" line ever follows, and the request stays in pending indefinitely. Manually clicking "Retry Request" just resets status without re-triggering a search (per the app's own shelfarr-spec.md, RequestQueueJob only re-queues status; actual searching is meant to happen via a recurring 5-minute background sweep). That sweep never appears to run either.

**Core evidence — `solid_queue_processes` is permanently empty:**
```
$ docker exec shelfarr sqlite3 /rails/storage/production_queue.sqlite3 "SELECT * FROM solid_queue_processes;"
(no output)
```
This is true from first container boot and persists across multiple full container restarts. No Solid Queue supervisor/worker/dispatcher startup line ever appears anywhere in container logs — only Puma's own boot sequence.

**Enqueue succeeds, execution never happens — reproduced twice:**
```
[ActiveJob] Enqueued SearchJob (Job ID: b47ddd27-ea5d-49f9-b36f-e2708a60ebcc) to SolidQueue(default) with arguments: 2
```

No further log activity for this job ever appears — no "Performing," no error, no exception, nothing. Reproduced both by clicking Retry on an existing stuck request, and by creating a brand-new request with "Immediate Search Enabled" turned on (which per its own description should search instantly on creation).

**Ruled out as causes:**

- Prowlarr connectivity — confirmed instant, correct response via curl from inside the container directly against Prowlarr's health endpoint.
- Missing queue database — production_queue.sqlite3 exists alongside its -wal/-shm files and shows active write activity.
- A one-off stuck worker needing a restart — a full docker restart does not change anything; the queue table is still empty afterward.
- Missing app credentials blocking worker boot — after separately identifying and fixing two missing-environment-variable issues (ACTIVE_RECORD_ENCRYPTION_* keys and SECRET_KEY_BASE, filed separately), bin/rails runner now executes cleanly with no crash — but solid_queue_processes is still empty and SearchJob still never runs. Since Puma's own process was clearly booting and serving requests fine the entire time (meaning it already had working credentials in its own environment), a worker forked from that same process should inherit the same working environment — so this doesn't appear to be the root cause of the main issue, just a separate problem encountered along the way.

**Confirmed correct — ruled out:** the actual puma.rb in the running image has the expected, standard pattern: 
`plugin :solid_queue if ENV["SOLID_QUEUE_IN_PUMA"]`
`SOLID_QUEUE_IN_PUMA` is confirmed set as a real container env var, so this condition evaluates true and the plugin line is reached. This rules out a misconfigured flag as the cause.

Given the plugin is confirmed to load, `SolidQueue::Supervisor.start` appears to fail or exit silently on boot — no error, no exception, no startup banner, nothing in the logs at all to indicate it ever ran. Happy to add any specific debug output or Solid Queue verbosity flag if there's one you'd recommend checking.

Pedro-Revez-Silva:
## Assigned issue

Closes #517

## What changed

On the reported Docker path, searches were enqueued while no Solid Queue processes registered. With `SOLID_QUEUE_IN_PUMA=1`, Puma now spawns a fresh `bin/jobs` process instead of using the upstream plugin's post-boot fork.

Only Puma's master starts the supervisor. Web worker recycling cannot stop it, stale pidfiles are never adopted, and shutdown/restart stop the owned supervisor with bounded cleanup. Unexpected supervisor exit stops Puma so the deployment's restart policy can recover; the queue also monitors its parent. Startup and shutdown remain safe when a signal arrives during spawning.

## Verification

- Real Puma probes cover single-process mode, clustered mode with and without preloading, exactly one supervisor, completed `SearchJob`, stale pidfiles, worker replacement, hot restart, graceful shutdown, supervisor failure, and abrupt parent exit.
- A deterministic shutdown-during-spawn regression reproduced a deadlock before the fix and passed afterward.
- `CI=true RAILS_ENV=test PARALLEL_WORKERS=4 bin/quality push`: RuboCop, Brakeman, Rails tests, system tests, and coverage.
- Built and booted the production Docker image through its normal entrypoint and Thruster, using jemalloc, PUID=99/PGID=100, and bind-mounted SQLite storage. A `SearchJob` completed with exactly one supervisor; container restart restored one supervisor.
- Independent adversarial review and re-review completed. No UI changes.

The reporter's actual Unraid host/filesystem was not available; the matching container settings were exercised locally.


Pedro-Revez-Silva (commented):
Adversarial review complete at head `9a0b17343394958cab5041612d52083901ab32b5`, against merge-base `e1f5f37dbc513a138fe2aed88a7e885b84f57213`.

Ready to merge. The review found and fixed lifecycle defects in the original approach: Rails initializer startup could create extra supervisors in cluster workers; shutdown/restart hooks did not match the resolved Puma API; startup interrupted by shutdown could deadlock; and an abruptly terminated Puma master could leave the queue supervisor behind. The final implementation starts a fresh `bin/jobs` process only from Puma's master boot hook, owns and reaps that child, and couples their lifetimes with bounded cleanup.

Verification included independent runtime and validation reviews; real Puma single, clustered, and preloaded execution; a real SearchJob reaching completion; stale pidfile handling; worker recycling; hot restart; queue failure; graceful shutdown; abrupt parent death; and deterministic shutdown-during-startup regression coverage. The focused lifecycle suite passes 11 tests / 40 assertions. Full local lint, Brakeman, test, system-test, and coverage gates passed. The final PR CI checks passed, including filesystem contracts on x86-64 and ARM and the container build.

A production image was also booted through the default Docker entrypoint and Thruster with a bind-mounted storage directory and PUID=99/PGID=100. A real SearchJob completed, the expected supervisor/worker/dispatcher/scheduler registered, and a container restart restored one supervisor. An actual Unraid filesystem was not available, so that host-specific environment remains unverified.

No blocking findings remain. GitHub does not permit an account to approve its own PR; this comment records the positive readiness decision, with the maintainer's explicit authorization to merge after verification.

PR #531, opened 2026-09-10
Fix Libation companion healthcheck launching a second server

### LINKED ISSUE #530: Libation companion healthcheck launches a second server and reports the container unhealthy
## Summary

The `shelfarr-libation` container remains `unhealthy` even though the Libation companion process is running and serving its HTTP API successfully.

The image healthcheck appears to launch a second instance of the companion server instead of checking the already-running instance. The second instance fails because port `8282` is already in use.

## Environment

- Shelfarr image: `ghcr.io/pedro-revez-silva/shelfarr:latest`
- Libation companion image: `ghcr.io/pedro-revez-silva/shelfarr-libation:latest`
- Companion version: `2026.08.31.1`
- Libation version: `13.5.1`
- Docker Compose under WSL2 on Windows Server 2022
- Companion URL: `http://0.0.0.0:8282`

## Current healthcheck

The image config contains:

```yaml
healthcheck:
  test: ["CMD", "/companion/Shelfarr.Libation.Companion", "--healthcheck"]
```

## Observed behavior

The container process is running continuously with no restarts:

```text
shelfarr-libation   Up 7 days (unhealthy)
state=running health=unhealthy restarts=0
```

The configured healthcheck exits with code `1` and produces no output:

```text
healthcheck_exit=1
```

Running the companion command manually produces an address-binding failure:

```text
Hosting failed to start
System.IO.IOException: Failed to bind to address http://0.0.0.0:8282: address already in use.
```

The existing companion process is listening on port `8282`, and the `/health` endpoint responds successfully:

```http
GET /health HTTP/1.1

200 OK
{"status":"ok","apiVersion":"1","companionVersion":"2026.08.31.1","libationVersion":"13.5.1","libraryReady":false,"busy":false}
```

`libraryReady:false` is expected for an installation where Libation is not configured yet. It does not indicate that the HTTP companion process is unavailable.

## Relationship to an unconfigured Libation installation

This installation has the Libation companion enabled, but Libation itself has not been configured with an account or library. Inside the container, the initial configuration files are empty or contain only empty JSON objects:

```text
/config/AccountsSettings.json  {}
/config/Settings.json          {}
/config/LibationContext.db     0 bytes
```

The companion still starts and exposes its API in this state. Its `/health` response uses `libraryReady:false` to report that Libation is not ready to process audiobook jobs, while returning HTTP `200` to indicate that the companion service itself is alive.

The current Docker healthcheck does not appear to handle this distinction. It invokes the companion executable with `--healthcheck`, but that invocation starts another server process rather than querying the existing one. The second process fails at the socket-binding stage because the running companion already owns port `8282`, so Docker records a failed healthcheck before any useful readiness result can be returned.

Therefore, the unconfigured Libation state explains why `libraryReady` is `false`, but it should not make the companion container unhealthy. The healthcheck should test liveness separately from optional Libation/library readiness, since Shelfarr is expected to run with the companion installed but idle until Libation is configured.

## Expected behavior

The healthcheck should report healthy when the already-running companion is serving its health endpoint, including when Libation is installed but not configured.

## Possible fix

Consider changing the image healthcheck so it probes the existing HTTP service rather than launching another server process. For example, if an HTTP client is available in the image:

```dockerfile
HEALTHCHECK CMD curl --fail http://localhost:8282/health || exit 1
```

If the image intentionally does not include an HTTP client, the companion could provide a non-server healthcheck mode that checks the running service or its state without attempting to bind port `8282`.

## Workaround

The Compose-level workaround is to check that the existing process has opened port `8282` instead of invoking a second companion process:

```yaml
healthcheck:
  test: ["CMD-SHELL", "test -r /proc/1/cmdline && grep -q ':205A ' /proc/net/tcp"]
```

Pedro-Revez-Silva:
## Assigned issue

Closes #530

## What changed

The `shelfarr-libation` image healthcheck already invoked
`/companion/Shelfarr.Libation.Companion --healthcheck`, but that path was easy
to miss and always probed `http://127.0.0.1:8080/health`.

If Docker, the host, or `ASPNETCORE_URLS` supplied extra arguments or a
non-default listen port (the reporter used `8282`), `--healthcheck` could fall
through into a second Kestrel start and fail with "address already in use", or
probe the wrong port and exit `1` with no useful signal. `libraryReady: false`
on a live `GET /health` was also easy to treat as unreadiness even though that
field only means Libation is idle or unconfigured.

This change:

- Treats any `--healthcheck` argument as a probe and never starts Kestrel.
- Resolves the live listen address from `--urls`, then `DOTNET_URLS` (which
  overrides the image `ASPNETCORE_URLS` under `WebApplication.CreateBuilder`),
  then `ASPNETCORE_URLS` / `ASPNETCORE_HTTP_PORTS`.
- Probes wildcard/unspecified hosts on loopback and preserves concrete IPv4
  and IPv6 bind addresses (for example `127.0.0.2`).
- Treats HTTP 2xx with `status: ok` as live and ignores `libraryReady`.
- Leaves the entrypoint without argv forwarding so a container `command`
  `--urls` override cannot desync the running process from the image
  healthcheck. Remap the listen port with `ASPNETCORE_URLS` or `DOTNET_URLS`.
- Extends unit tests so the probe must talk to an already-bound listener and
  succeed when `libraryReady` is false.
- Exercises the live `--healthcheck` binary against the running smoke container.

## Verification

- `dotnet format --verify-no-changes` on the companion project and test project.
- `dotnet test` for `Shelfarr.Libation.Companion.Tests`: 108 passed, 0 failed.
- New `CompanionHealthProbeTests` cover:
  - no second bind of the listen port
  - liveness success when `libraryReady` is false
  - listen-port resolution from `ASPNETCORE_URLS` / `--urls` (including 8282)
  - `DOTNET_URLS` taking precedence over `ASPNETCORE_URLS`
  - a live listener bound to `127.0.0.2`
- GitHub CI `companion`, `quality`, and `security` checks passed on earlier
  revisions; container smoke is part of the companion job.

## Environment note

The original report used Docker Compose under WSL2 on Windows Server 2022 with
companion `2026.08.31.1` / Libation `13.5.1` and `ASPNETCORE_URLS` on port
`8282`. That exact host was not reproduced 1:1. The fix is covered by mocked
and in-process listener tests plus the Linux container smoke path.

## Screenshots or recordings

Not applicable; no visible or interactive UI change.

## Checklist

- [x] The maintainer assigned the linked issue before implementation started
      (assigned to Pedro-Revez-Silva)
- [x] This pull request stays within the assigned issue's scope
- [x] I added or updated tests for the behavior I changed
- [x] I ran the relevant local quality checks
- [x] I updated documentation for user-visible changes

<div><a href="https://cursor.com/agents/bc-de8e2e39-ac90-40a0-a809-0b70b8d49247?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-de8e2e39-ac90-40a0-a809-0b70b8d49247&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


services/libation_companion/src/Shelfarr.Libation.Companion/CompanionHealthProbe.cs
```diff
@@ -1,28 +1,210 @@
...
+        IReadOnlyDictionary<string, string?>? environment,
+        out string urls)
+    {
+        if (TryUrlsFromArguments(arguments, out var fromArguments))
+        {
+            urls = fromArguments;
+            return true;
+        }
+
+        var fromEnvironment = ReadVariable(environment, "ASPNETCORE_URLS")
```
Pedro-Revez-Silva: [P2] Match WebApplication's URL environment precedence

The image sets ASPNETCORE_URLS to port 8080. If an operator sets DOTNET_URLS to port 8282, WebApplication.CreateBuilder actually listens on 8282, but this fallback selects 8080 and reports the live container unhealthy. Reproduced with the pinned .NET SDK and a failing resolver regression (expected 8282, actual 8080). Prefer DOTNET_URLS before ASPNETCORE_URLS, matching the server configuration.

services/libation_companion/src/Shelfarr.Libation.Companion/CompanionHealthProbe.cs
```diff
@@ -1,28 +1,210 @@
...
+        var normalized = NormalizeKestrelBinding(binding);
+        if (!Uri.TryCreate(normalized, UriKind.Absolute, out var parsed)
+            || !string.Equals(parsed.Scheme, Uri.UriSchemeHttp, StringComparison.OrdinalIgnoreCase))
+            return false;
+
+        var port = parsed.IsDefaultPort ? 80 : parsed.Port;
+        if (port is < 1 or > 65535)
+            return false;
+
+        healthUri = LoopbackHealthUri(port, ipv6: parsed.HostNameType == UriHostNameType.IPv6);
```
Pedro-Revez-Silva: [P2] Preserve explicitly bound IP addresses in the probe

A valid ASPNETCORE_URLS=http://127.0.0.2:8282 binds Kestrel specifically to that address, but this line probes 127.0.0.1:8282. A live TcpListener regression bound to 127.0.0.2 reproduces exit 1; the wildcard case succeeds. Preserve concrete IPv4/IPv6 bindings and normalize only wildcard/unspecified hosts to loopback.

Pedro-Revez-Silva (commented):
Adversarial review found two reproducible healthcheck defects that need correction before merge. Reviewed head d878f8b3af4244880ae9f28bc51b5c3e0bc2d5bc, merge-base 726d9dbfe7a1fa53ecba9b7494fd1097f204f49c. Issue #530's default/custom-port liveness scope is otherwise covered by the early probe return, libraryReady handling and live-container smoke.

The original 107 tests passed. Added adversarial regressions produce two failures: incorrect DOTNET_URLS precedence and a failed probe of a live listener bound to 127.0.0.2. A separate .NET 10.0.302 WebApplication.CreateBuilder harness confirms DOTNET_URLS overrides ASPNETCORE_URLS. I will fix both and repeat verification under the maintainer's authorization.

Recorded as COMMENT because the authenticated account is the PR author and cannot request changes on its own PR. These are blocking review findings, not optional suggestions.

Pedro-Revez-Silva (commented):
Adversarial re-review complete: both verified findings are fixed; no actionable findings remain. Ready to merge.

Reviewed final head 6b60abb77c16b3ee92fe194539529b2c0d4170bc, merge-base 726d9dbfe7a1fa53ecba9b7494fd1097f204f49c. The concurrent author update was preserved and reviewed rather than overwritten by the independently prepared fixes. DOTNET_URLS now has the same precedence as WebApplication.CreateBuilder, concrete IP bindings are retained, and the probe returns before server construction. This addresses issue #530's incorrect custom-port health result and keeps libraryReady informational.

Exact-head verification: locked restore, both formatting checks and all 108 companion tests passed with the pinned .NET 10.0.302 SDK. The full Docker smoke suite passed. Two additional actual containers reported healthy when DOTNET_URLS selected port 8282 over the image default and when ASPNETCORE_URLS bound specifically to 127.0.0.2:8282; direct probe invocations exited successfully. The original failures had been reproduced before correction. Independent final runtime and deployment reviews found no further issues.

GitHub CI quality, gem security, companion tests/smoke, x86/ARM filesystem contracts, container build and CodeQL all passed at this exact head. The original WSL2 host was not reproduced; runtime verification used Linux containers. Temporary live containers were stopped.

Recorded as COMMENT because GitHub prohibits the authenticated PR author from self-approval; the review conclusion is approval for merge.

